PR #17, opened 2026-04-18
docs: fix full-text search example in welcome note

### LINKED ISSUE #16: Incorrect command for Full-text search example in Welcome Note
`- **Full-text search** — press ⌘K from anywhere`

I think you meant `⌘p`, but even that isn't full-text search. So maybe `<leader> st`?

samueljoli:
Fixes #16


adibhanna:
Thanks. The underlying issue is real, but this PR still keeps the incorrect `⌘K` guidance alive by mixing it with `<leader> st`. I patched the welcome note directly in the main branch here to say `Space s t` in Vim mode and point non-Vim users to **Search Text in Vault** from the command palette. Closing this as superseded by the direct fix. Note: that patch is in the branch here but not pushed yet.

PR #19, opened 2026-04-18
fix(linux): set SUID permissions for chrome-sandbox in post-install

### LINKED ISSUE #18: [Bug][Linux] App fails to launch in linux due to incorrect chrome-sandbox permissions
I found that after installing the .deb package (v1.0.2) on Ubuntu, the application fails to launch from the GUI. When attempted via terminal, the app crashes because the chrome-sandbox helper lacks the necessary SUID permissions required by modern Linux kernels.

Steps to Reproduce:

1. Download and install ZenNotes-1.0.2-linux-amd64.deb on Ubuntu/Debian.
2. Attempt to open the application from the app drawer.
3. Observe that the process starts and immediately terminates.

The chrome-sandbox binary in /opt/ZenNotes/ is not owned by root or does not have 4755 permissions set during the packaging process.

Manually running sudo chown root:root /opt/ZenNotes/chrome-sandbox && sudo chmod 4755 /opt/ZenNotes/chrome-sandbox resolves the issue and allows the app to launch normally. This should be automated in the afterInstall build hook.

I know the exact issue and would like to work on this. Waiting for your "Yes"

madesh02104:
Closes #18

### DIFF
--- a/build/after-install.sh
+++ b/build/after-install.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+# Fix SUID permissions for chrome-sandbox on Linux
+chown root:root /opt/ZenNotes/chrome-sandbox
+chmod 4755 /opt/ZenNotes/chrome-sandbox
--- a/package.json
+++ b/package.json
@@ -149,6 +149,9 @@
       ],
       "maintainer": "Adib Hanna <adibhanna@gmail.com>",
       "category": "Office"
+    },
+    "deb": {
+      "afterInstall": "build/after-install.sh"
     }
   }
 }


adibhanna:
Thanks for the report and patch. The Debian issue is valid, but I’m not merging this exact version because the post-install script hard-fails if /opt/ZenNotes/chrome-sandbox is missing or the package layout changes, which can break installs/upgrades. I’ve reworked the Debian fix locally as a guarded afterInstall script that exits cleanly when the helper is not present, and I’m keeping AppImage out of this PR because forcing a no-sandbox workaround there would be the wrong default security tradeoff. Closing this PR so we don’t merge the brittle version as-is.

adibhanna:
Closing this in favor of a hardened follow-up to the Debian packaging fix.

madesh02104:
Cool! that worked. I will start using the app then :)

Issue #63, opened 2026-06-07
[Feat] Calendar panel in sidebar for daily/weekly notes


10wpressure:
### Description:
When using daily or weekly notes there's no quick way to jump between days or see which ones have notes written.

Would love a calendar panel in the sidebar (similar to the obsidian-calendar-plugin, screenshot below) - clicking a day opens/creates the daily note for that day, clicking a week number opens/creates the weekly note. The panel would only appear when daily or weekly notes are enabled in vault settings.

Also we would have a global toggle to enable/disable calendar.

I already have an implementation working locally, happy to open a PR if the
approach sounds good.

---
### Obsidian

<img width="536" height="356" alt="Image" src="https://github.com/user-attachments/assets/1569e66b-58ae-41b7-8bd2-588a40c8ba9a" />

---

### ZenNotes
<img width="322" height="492" alt="Image" src="https://github.com/user-attachments/assets/58919578-97a8-4d81-95a0-3cff07c0c1ac" /> 

<img width="1119" height="982" alt="Image" src="https://github.com/user-attachments/assets/ae73cba7-e4ee-4996-b092-694549efa47e" />

adibhanna:
Shipped 🎉 — thanks for the great writeup and the reference shots, @10wpressure.

We implemented calendar navigation for daily/weekly notes, with a slightly different design than the left-sidebar panel:

- **Right-side calendar panel** that auto-appears when the active note is a daily (`YYYY-MM-DD`) or weekly (`YYYY-Www`) note, anchored to and highlighting that note. Toggle it anytime with the calendar icon, the command palette, or `leader c`; it stays available on any note when daily/weekly notes are enabled.
- Click a day to open/create its daily note; click a week number for the weekly note; page months with the chevrons / "Today".
- **Obsidian-parity touches:** word-count dots (more writing → more dots), a marker for days with unfinished tasks, hover preview, right-click open/create/move-to-trash, configurable **week start** (Mon/Sun/locale) and a **show week numbers** toggle, plus a global enable/disable setting.
- **Sidebar:** Daily/Weekly directories are also surfaced as pinned, date-grouped sections (year → month → day) for jumping between dates.

This will ship in the **next release**. Closing as resolved-pending-release.

PR #66, opened 2026-06-07
feat: calendar sidebar panel for daily and weekly notes

10wpressure:
Adds a collapsible calendar panel to the sidebar for navigating daily and weekly notes. Clicking a day opens or creates the daily note for that date; clicking a week number does the same for weekly notes. The panel only appears when daily or weekly notes are enabled, and can be turned off in Settings -> Calendar.

Ref #63

### DIFF
--- a/packages/app-core/src/components/DailyNotesCalendar.tsx
+++ b/packages/app-core/src/components/DailyNotesCalendar.tsx
@@ -1,0 +1,243 @@
+import { useCallback, useMemo, useState } from 'react'
+import { useStore } from '../store'
+import {
+  normalizeVaultSettings,
+  noteFolderSubpath,
+  weeklyNoteTitle,
+} from '../lib/vault-layout'
+import { getISOWeek, getISOWeekYear } from '../lib/template-render'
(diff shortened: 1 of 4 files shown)


adibhanna:
Thank you for this, @10wpressure — both the [#63](https://github.com/ZenNotes/zennotes/issues/63) proposal and this implementation kicked the whole feature off. 🙏

We ended up taking the calendar in a different direction, so this is landing as a separate implementation rather than a merge of this branch:

- A **right-side, contextual** calendar panel (auto-opens on daily/weekly notes) instead of an always-on left-sidebar panel.
- Extra Obsidian-parity behavior — word-count dots, an open-task marker, hover preview, right-click open/create/trash, configurable week start + week numbers.
- The Daily/Weekly directories are surfaced in the left sidebar as pinned, **year → month → day** grouped sections so the list stays compact even with lots of notes.

Because the approach diverged this far, I'm **closing this PR** — but the feature you asked for is in and ships in the **next release**. Really appreciate you getting the ball rolling, and sorry we didn't merge your branch directly. Definitely keep the contributions coming!

PR #95, opened 2026-06-10
Fix: set world-readable permissions on /opt/zennotes-bin after extraction. 

bitwisemonkk:
the `cp -a` call preserves the permissions of the extracted squashfs-root
directory, which comes out as `drwx------` (root-only). This means non-root
users cannot enter /opt/zennotes-bin/ at all, causing `zennotes: permission
denied` on launch even after the symlink in /usr/bin is correctly set.

Fix: add `chmod -R a+rX` after the cp to ensure directories are traversable
and files are readable by all users. The `X` (capital) flag only sets execute
on directories and already-executable files, so regular data files are not
made executable.


bitwisemonkk:
closing as this appears to already be fixed in main lol.

Issue #70, opened 2026-06-08
zennotes-bin installs /opt/zennotes-bin with restrictive permissions, breaking desktop launcher


lpanebr:
I installed `zennotes-bin` through Omarchy's `Install > AUR` flow. The package installed successfully and created the desktop entry:

```text
/usr/share/applications/zennotes.desktop
```

But launching ZenNotes from the app launcher failed with:

```text
App failure
Error: Command not found: "zennotes"
```

The desktop file uses:

```ini
Exec=zennotes %U
```

`/usr/bin/zennotes` existed and pointed to:

```text
/usr/bin/zennotes -> /opt/zennotes-bin/AppRun
```

However, `/opt/zennotes-bin` was installed with restrictive permissions:

```text
drwx------ root root /opt/zennotes-bin
```

Because regular users cannot traverse that directory, `/usr/bin/zennotes` is not usable from the desktop launcher or shell.

Running the following fixed it locally:

```bash
sudo chmod 755 /opt/zennotes-bin
sudo find /opt/zennotes-bin -type d -exec chmod 755 {} +
sudo chmod 755 /opt/zennotes-bin/AppRun /opt/zennotes-bin/ZenNotes
```

Expected behavior:

`zennotes-bin` should install `/opt/zennotes-bin` and its required subdirectories with permissions that allow regular users to execute `/usr/bin/zennotes`, e.g. directories mode `755` and
`AppRun`/`ZenNotes` mode `755`.

**Disclosure:**

I used Codex to help investigate, understand, and draft this report. I personally reviewed and tested the issue, the local workaround, and this issue description before submitting.

adibhanna:
Thanks for the precise report — the `drwx------ root root` on `/opt/zennotes-bin` was exactly it.

**Root cause:** the PKGBUILD installed the extracted AppImage with `cp -a squashfs-root/. /opt/zennotes-bin/`, and `cp -a` re-applied the squashfs root's restrictive mode (`0700`) onto `/opt/zennotes-bin`. With that directory not traversable by non-root users, the `/usr/bin/zennotes -> /opt/zennotes-bin/AppRun` symlink target couldn't be reached, so the launcher reported "Command not found".

**Fix** (9fde35f): normalize the installed tree with `chmod -R a+rX` so directories are world-traversable and files world-readable, while preserving the `AppRun`/`chrome-sandbox` exec + setuid bits. Bumped to `zennotes-bin 2.1.0-2`.

[...]

PR #133, opened 2026-06-12
feat: add tar.gz Linux target and simplify AUR PKGBUILD

### LINKED ISSUE #103: Publish a plain Linux tar.gz for cleaner Arch packaging
The `.pacman` release asset installs fine with `pacman -U`, but sideloaded
packages get no update tracking so on Arch the practical channel is the AUR
package you already maintain. Suggestion to make that package simpler and more
robust:

Add `"tar.gz"` to `linux.target` in the electron-builder config (it's a
supported archive target) and switch `zennotes-bin`'s `source=` from the
AppImage to that tarball. A plain tarball extracts with correct permissions,
so the `--appimage-extract` chmod workarounds in the PKGBUILD
could be deleted. This is the same pattern Vicinae uses: upstream publishes a
generic Linux tar.gz, and the AUR `-bin` package repackages it.

Lund1337:
- Add tar.gz to electron-builder Linux targets
- Include desktop file (copy of the appimage), icons via arch-extras in extraResources
- Include LICENSE via extraFiles
- Rewrite PKGBUILD to use tar.gz instead of AppImage extraction
- Fixes #103

## Verified

Tarball produced by `npm run dist:linux` (`ZenNotes-2.3.0-linux-x64.tar.gz`):

### Confirmed resources/arch-extras exists with desktop file + icons
```sh
tar -tzf ZenNotes-2.3.0-linux-x64.tar.gz | grep -i arch-extras
```

### LICENSE present at tarball root
```sh
tar -tzf ZenNotes-2.3.0-linux-x64.tar.gz | grep -i license
```

### chrome_crashpad_handler ships with correct execute permissions (-rwxr-xr-x) 
```sh
tar -tvzf ZenNotes-2.3.0-linux-x64.tar.gz | grep crashpad
```

**Note:** sha256sums in `.SRCINFO` / `PKGBUILD` still reference the old AppImage hash.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+github: adibhanna
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ On macOS, the first-party `zen` CLI also powers launcher workflows such as the R
 
 Grab the latest build from [GitHub Releases](https://github.com/ZenNotes/zennotes/releases/latest) — see [Install](#install) below.  
 Website: [zennotes.org](https://zennotes.org)
+Support development: [GitHub Sponsors](https://github.com/sponsors/adibhanna)
 
 Detailed in-repo documentation lives under [docs/README.md](docs/README.md).
 
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -152,20 +152,6 @@
           "!coverage/**",
           "!*.log"
         ]
-      },
-      {
-        "from": "../../packaging/aur/arch-extras/zennotes.desktop",
-        "to": "arch-extras/zennotes.desktop"
-      },
-      {
-        "from": "build/icons",
-        "to": "arch-extras/icons"
-      }
-    ],
-    "extraFiles": [
-      {
-        "from": "../../LICENSE",
-        "to": "LICENSE"
       }
     ],
     "afterSign": "scripts/notarize.cjs",
@@ -246,7 +232,23 @@
       ],
       "icon": "build/icons",
       "maintainer": "Adib Hanna <adibhanna@gmail.com>",
-      "category": "Office"
+      "category": "Office",
+      "extraResources": [
+        {
+          "from": "../../packaging/aur/arch-extras/zennotes.desktop",
+          "to": "arch-extras/zennotes.desktop"
+        },
+        {
+          "from": "build/icons",
+          "to": "arch-extras/icons"
+        }
+      ],
+      "extraFiles": [
+        {
+          "from": "../../LICENSE",
+          "to": "LICENSE"
+        }
+      ]
     },
     "deb": {
       "afterInstall": "build/after-install.sh"
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -152,6 +152,20 @@
           "!coverage/**",
           "!*.log"
         ]
+      },
+      {
+        "from": "../../packaging/aur/arch-extras/zennotes.desktop",
+        "to": "arch-extras/zennotes.desktop"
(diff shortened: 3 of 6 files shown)


adibhanna (changes requested):
Nice direction — moving the AUR `-bin` package off AppImage extraction onto a plain tarball is the right call, and dropping the whole `--appimage-extract` + chmod dance (#70/#74/#92) is justified since tar preserves permissions. I verified the two load-bearing assumptions and they hold:

- **Top-level tar dir** — electron-builder's `tar()` sets `prefix: path.basename(outFile, ".tar.gz")`, so the archive's top folder is exactly `ZenNotes-<ver>-linux-x64`, matching the PKGBUILD's `_extracted`. ✅
- **Symlink target** — `executableName: "ZenNotes"` means the binary is named `ZenNotes`, so `/usr/bin/zennotes → /opt/zennotes-bin/ZenNotes` and `Exec=zennotes` are correct. ✅

A few blocking items before merge:

### 1. Scope the new `extraResources`/`extraFiles` to `linux`
`extraFiles` (LICENSE) and the new `arch-extras` `extraResources` entries are added at the **global `build`** level, so they ship in the mac and win builds too. On macOS `extraFiles` copies into `ZenNotes.app/Contents/`, leaving a stray `Contents/LICENSE` at the bundle root — and this app notarizes (`afterSign: scripts/notarize.cjs`), where unexpected top-level files in `Contents/` are a common signing/notarization failure. The Linux `.desktop` + 8 PNG icons inside the mac/win installers are also pure bloat. Please move both new blocks under `"linux"` (electron-builder supports per-platform `extraFiles`/`extraResources`).

### 2. `pkgver` 2.3.0 points at a tarball that doesn't exist
The v2.3.0 release only has `.AppImage`, `.deb`, and `.pacman` — the `tar.gz` is produced by *this PR*, so `…/releases/download/v2.3.0/ZenNotes-2.3.0-linux-x64.tar.gz` 404s and `makepkg` can't fetch the source. The PKGBUILD/.SRCINFO need to target the first release that actually ships the tarball (the next release), not 2.3.0.

### 3. `sha256sums` still reference the old AppImage hash
As you noted, `f7f6a67b…` is the v2.2.0 AppImage hash. Needs regenerating with `updpkgsums` against the real published tarball — gated on #2, since the asset has to exist first.

### Minor
- You verified `chrome_crashpad_handler` is `-rwxr-xr-x`; the main `ZenNotes` binary's exec bit is equally load-bearing for launch — worth a quick `tar -tvzf … | grep -E '/ZenNotes$'` to confirm.
- The removed PKGBUILD comments captured real context (#70/#74/#92). A one-liner noting "tar.gz preserves perms, so the chmod workarounds are gone" would keep that history.

Lund1337:
Addressed review feedback:

1. **Scoped to linux** — moved `extraResources` (arch-extras) and `extraFiles` (LICENSE) from top-level `build` into the `linux` block
2. **pkgver bumped to 2.4.0** — the tar.gz will ship with the next release, not 2.3.0
3. **Added comment in PKGBUILD** — 
```
# tar.gz preserves permissions (unlike squashfs extraction)
# chmod workarounds from #70/#74/#92 are no longer needed.
```
4. **Verified ZenNotes binary permissions** — `-rwxr-xr-x 0/0       206339288 2026-06-12 16:30 ZenNotes-2.3.0-linux-x64/ZenNotes` with command:
```sh
tar -tvzf ZenNotes-2.3.0-linux-x64.tar.gz | grep -E '/ZenNotes$'   
```


sha256sums still needs `updpkgsums` when the release is published — gated on the release shipping the tar.gz artifact.

PR #140, opened 2026-06-12
feat(ui): add middle-click to close tabs

victor-rm:
This PR introduces a small UX improvement by adding the ability to close tabs using the middle mouse button.

Since this is the standard behavior for tabs in most modern web browsers, it makes navigating the editor pane feel much more natural even for me as a heavy vim motion user. Users can now quickly close tabs without needing to aim exactly at the small close icon.

It's a very straightforward and isolated change, so I decided to open a PR directly for it.

I'm in love with the app! Thank you.

### DIFF
--- a/packages/app-core/src/components/EditorPane.tsx
+++ b/packages/app-core/src/components/EditorPane.tsx
@@ -2453,6 +2453,17 @@ export function EditorPane({ pane }: { pane: PaneLeaf }): JSX.Element {
             e.preventDefault()
             setTabMenu({ x: e.clientX, y: e.clientY, path: tab.path })
           }}
+          onMouseDown={(e) => {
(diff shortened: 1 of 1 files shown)


adibhanna:
Thank you, this is a great little addition — clean, well-scoped, and the right implementation (`onAuxClick` for button 1 with an `onMouseDown` `preventDefault` to suppress the middle-click autoscroll/paste). Exactly how browsers do it.

I cherry-picked your commit onto the active **`release/v2.4.0`** branch (`4cdbb22`) **with your authorship preserved**, so it'll ship in **v2.4.0**. Closing this PR (which targeted `main`) since the change is already in the release branch — not because it was rejected. 🙏

So glad you're enjoying the app, and thanks for contributing!

Issue #143, opened 2026-06-12
.vimrc support


christphralden:
Very excited on this project first of all

# Description

Personally i remap my keys for a boost of productivity, but currently zennotes only supports limited rebinds for vimmotions.
I tried to edit keymaps the best i possibly can, but not all motions are supported in the remaps

adding a .vimrc support over the existing keymap for vim would be gamechanger for many users

christphralden:
@adibhanna id love to contribute to this project, and as per CONTRIBUTING.md if theres any discussion needed, feel free to scope it out. 

but for personal use i forked and created my own implementation on this for now. happy to align

adibhanna:
thank you for your comment! I definitely welcome your contribution! please open a PR for this one! I know I need to improve vim support even more, but im currently the only dev working on this app

PR #146, opened 2026-06-13
feature(vimscript-mappings): add support to extend base keymaps with vimscript

christphralden:
Issue https://github.com/ZenNotes/zennotes/issues/143
# Goal
Vimscript support to enable flexible remapping

# Design desicion
Currently codebase has contracts for keymaps with different scopes of `KeymapScope`. What this PR aims to do is not do a full refactor of existing architecture, but to allow flexibility for users to define their own keymap

> User defined vimscript will be referred as `user-mappings` for this context

user-mappings are treated as a different entity from `KeymapDefinition`. Why? 
1. To eliminate complexity early on from managed keys and user's preferences
2. Allows complete flexibility to users as long as they know what they are doing (which if you use this app, you probably used vim before)

user-mappings interact directly with codemirror's api

user-mappings are defined inside settings > keymaps > Vimrc opposed as to source it using the file system to give reactive feedback like conflicting keymaps and errors. This is also because we could not possibly support every syntax for example setting a variable.

Going with this direction, to improve UX, conflicting keymaps should be surfaced as warnings for users to manage

# Summary
- Add Vimscript support for custom key mappings in the Editor settings, surfaced as a textarea in the Keymaps panel 
- Introduce diagnostics for user-mappings and flag unknown commands, missing RHS, and conflicts with existing app keymaps inline below the textarea
Supported syntax are as such
```ts
export type VimMode = 'normal' | 'visual' | 'insert'

export type VimMapCmd =
  | 'noremap'
  | 'map'
  | 'nnoremap'
  | 'nmap'
  | 'vnoremap'
  | 'vmap'
  | 'xnoremap'
  | 'xmap'
  | 'inoremap'
  | 'imap'

export type VimUnmapCmd = 'unmap' | 'nunmap' | 'vunmap' | 'xunmap' | 'iunmap'
```
-  Detect keymap collisions which is a plus feature

# Notes
There are noisy diffs from differences in linter, if you could specify a .prettierrc it would be great :)

---

image

### DIFF
--- a/packages/app-core/src/components/Editor.tsx
+++ b/packages/app-core/src/components/Editor.tsx
@@ -44,6 +44,7 @@ import { applyVimInsertEscape } from '../lib/vim-insert-escape'
 
 let vimCommandsRegistered = false
 let syncedVimBindings: Partial<Record<KeymapId, string[]>> = {}
+let appliedVimMappingLhs: string[] = []
 
 const DEFAULT_VIM_MAPPINGS_TO_CLEAR = [
   'gd',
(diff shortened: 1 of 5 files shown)


Tornado300:
i would love a way of adding my vim config into zennotes.

adibhanna:
Closing as stale: it has drifted into conflict against 44 days of changes and is a large diff. #143 stays open as the tracking issue; if you want to rebase a smaller version I will review it.

PR #162, opened 2026-06-14
add optional native Linux title bar setting

ptdewey:
Adds a Linux-only appearance setting to use the native window-manager title bar instead of the default electron title bar. This option will only show up in the settings page on a linux system, and requires a restartn to take effect (the electron window needs to be recreated to adjust window decorations). Default behavior remains the same as before.

Screenshots:

With default electron title bar:
image

No title bar (based on my window manager setting):

image

Toggle in settings:
image

Sorry if I should have opened an issue for this first, but I figured this would a relatively small and unobjectionable change.


wholetthekfcgo:
+1

radiorambo:
Fixes https://github.com/ZenNotes/zennotes/issues/263

raphael-soares:
+1

Kamyil:
omg, I would love to have this on macOS as well. Default title bar takes so much space, I would like to have something like Helium browser has

Tornado300:
yes please. that would be awesome!.

Tornado300:
for everyone that wants to have a quick fix/workaround for removing the title bar, you can just remove the whole bar with css.
```css
.glass-titlebar {
	display: none;
}
```

maybe that helps someone.

Kamyil:
@Tornado300 Holy cow, it's soooo sooo much better. Thank you so much!
image

Tornado300:
@Kamyil no problem, css overrides are powerful. 
PS: if you are on the discord, there is a theme channel where everyone can put css snippets. i have also put a snippet for better tabs. it makes them a bit bigger, rounder, changes the limiter bar  and some other stuff. im also doin some other snippets and might post them there if they are worth it.

PR #174, opened 2026-06-14
fix: set quick capture window title

### LINKED ISSUE #155: Quick capture window is indistinguishable from the main window on Wayland, breaking WM-side hotkey binding
**Platform:** Linux/Wayland
**Compositor:** Niri

### What's happening

The quick capture floating window shares the same `app_id` and window title title as the main ZenNotes window. That's expected - both windows come from the same app. The problem is that there's no secondary indentifier (e.g a distinct window title or something) to tell them apart.

On standalone Wayland compositors, global hotkeys are not registered by apps - they must be bound manually in the compositor config. To bind a key that opens/toggles the Quick Capture floating window specifically when the app has no focus, users  need something to match against. Right now there's nothing.

### Steps to reproduce (in Niri)

1.  Open ZenNotes
2. Trigger the Quick Capture floating window
3. Run a window inspection command, e.g`niri msg pick-window'

#### Output

 niri msg pick-window
Window ID 9: (focused)
  Title: "ZenNotes"
  App ID: "ZenNotes"
  Is floating: no
  PID: 180579
  Workspace ID: 4
  Layout:
    Tile size: 900 x 705
    Scrolling position: column 1, tile 1
    Window size: 900 x 705
    Window offset in tile: 0 x 0

### Expected behaviour

The quick capture window should have a distinct window title - for example `ZenNotes Quick Capture` - so users can match on it  in compositor rules:

```
window-rule {
  match title="ZenNotes Quick Capture"
  open-floating true
}
```

### Why this matters

The in-app system-wide hotkey registration doesn't work on standalone WMs, those environments intentionally leave hotkey management to the compositor. The only path forward for WM users is a manual bind, which requires something to target. Without a distinct window title, the quick capture feature is effectively unavailable outside of a full DE.

### Possible fix

Set a distinct, stable window title, something like`ZenNotes Quick Capture`. That's all that's needed for compositor rules to work.

dannrs:
## Summary

This small PR fixes the title of the quick capture window, making it distinguishable from the main `ZenNotes` window on Wayland compositors.

The quick capture window now has a different title: `ZenNotes Quick Capture`

This gives compositor rules something stable to match against.

Fixes [#155](https://github.com/ZenNotes/zennotes/issues/155)

## Testing:

- Open ZenNotes on Linux with a Wayland compositor. I tested with `niri`.
- Open the quick capture window.
- Run `niri msg pick-window` or an equivalent window inspection command.
- Confirm that the title of the quick capture window is now: `ZenNotes Quick Capture`.

### DIFF
--- a/apps/desktop/src/main/index.ts
+++ b/apps/desktop/src/main/index.ts
@@ -2755,6 +2755,7 @@ function ensureQuickCaptureWindow(): BrowserWindow {
     height: 340,
     minWidth: 460,
     minHeight: 260,
+    title: 'ZenNotes Quick Capture',
     show: false,
     frame: false,
     titleBarStyle: mac ? 'hiddenInset' : 'hidden',
--- a/packages/app-core/src/components/QuickCaptureApp.tsx
+++ b/packages/app-core/src/components/QuickCaptureApp.tsx
@@ -228,6 +228,11 @@ export function QuickCaptureApp(): JSX.Element {
   const [overlay, setOverlay] = useState<'none' | 'search' | 'command'>('none')
   const editorRef = useRef<EditorView | null>(null)
 
+  // Set a different title for the quick capture window.
+  useEffect(() => {
+    document.title = 'ZenNotes Quick Capture'
+  }, [])
+
   // Apply theme + font CSS vars before paint.
   useEffect(() => {
     applyTheme(prefs)


adibhanna:
Merged into the active release branch (\`release/v2.4.0\`) as b26395b rather than via this PR — your branch also carried two unrelated commits from \`main\` (the GitHub Sponsors / \`FUNDING.yml\` changes), so I cherry-picked just your fix commit onto the release branch to keep it clean. Your authorship is preserved on the commit. Nothing was rejected — this ships in v2.4.0.

Thanks @dannrs! Clean, correct fix: the constructor \`title\` plus \`document.title\` in \`QuickCaptureApp\` is exactly the right combo so the title sticks after the renderer loads, giving Wayland compositors (niri & co.) a stable \`ZenNotes Quick Capture\` toplevel to match against. Closes #155.

PR #149, opened 2026-06-13
feat(packaging): add nix packaging 

hallwack:
## Summary

Add Nix packaging for ZenNotes under `packaging/nix`.

The original implementation used the official AppImage release as its source and followed a similar approach to the existing AUR and Flatpak packaging by extracting the AppImage during the build process.

## Tested

- [x] Desktop package builds successfully
- [x] Server package builds successfully
- [x] Desktop startup on NixOS 25.05
- [x] Desktop entry generation
- [x] Icon installation
- [x] `nix profile install`

## Notes

The original implementation relied on the upstream AppImage release and was therefore limited to the platforms supported by the available AppImage artifacts.

## Update

Following review and feedback, the implementation has evolved beyond the original AppImage-based package. 

The current implementation now builds ZenNotes from source using Electron and includes several additional improvements:

* Separate desktop and server packages
* Shared release metadata via `packaging/nix/release-data.json`
* Flake outputs for desktop and server packages
* Optional CLI installation support
* Updated Nix packaging documentation

This approach is more closely aligned with common nixpkgs packaging practices, improves platform support, and removes the dependency on upstream AppImage releases.

Additional improvements were contributed through follow-up work by @justkrysteq and have been merged into this branch.

PR #176, opened 2026-06-14
chore: add a nix flake to the project root

grunisaurus:
**What this pr does:**
- Adds flake.nix & flake.lock: Sets up a deterministic environment for everyone using the nix package manager.
- devShells.default: A isolated shell with all the CLI tools, linters, and libraries pre-loaded.
- packages.default: A clean, hermetic way to build the project locally via Nix and embedd it into their system configuration.

**How to use:**
`nix develop` -> creates a new shell with all the packages defined for the environment

`nix build` -> runs through all the build steps defined in the flake.nix, builds it and creates a ./result symlink into the /nix/store to the built application.

**Steps:**
[x] `nix flake check` succeeds
[x] The application starts and runs after building
[x] Verified the Linux build
[ ] Verified the MacOS build

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1781074563,
+        "narHash": "sha256-md8WlXOlfnIeHeOScMTTHFyf2d6iaTwPl2apR5EQ3P4=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "9ae611a455b90cf061d8f332b977e387bda8e1ca",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "nixpkgs": "nixpkgs"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
--- a/flake.lock
+++ b/flake.lock
@@ -1,5 +1,23 @@
 {
   "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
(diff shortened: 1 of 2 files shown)


adibhanna:
Thank you for this, @grunisaurus — this is a solid, working flake and I appreciate the care that went into it (the `nix develop` shell, the darwin/linux split, the go-embed wiring for the server).

The timing is just unlucky: a Nix flake from #149 landed on our active release branch (`release/v2.4.0`) right before this, so the repo already has a root `flake.nix` plus a `packaging/nix/` layout. Since both add `flake.nix`/`flake.lock` at the same paths they can't coexist, and I don't want to churn the one that's already in. So I'm closing this as superseded — **not** rejected on quality.

That said, two things your flake does that the merged one doesn't, and that I'd genuinely love as a follow-up PR against the new `packaging/nix/` structure:

1. **A dev shell that builds from the local working tree** (`nix build` of the current checkout, not a pinned release tag) — great for contributors hacking on the app. The merged flake builds a published release via `fetchFromGitHub`, so a from-source dev path would complement it nicely.
2. **The macOS `.app` bundle** — the merged desktop package is Linux-focused, so a verified darwin build would be a real addition.

One heads-up if you pick up the desktop build: the `postFixup` here `sed`s the renderer-trust check (`isTrustedRendererUrl`) to allow `nix/store` URLs. That's touching a security-sensitive gate, so I'd want to find a cleaner way to make Electron load from the store than patching that function — happy to work through it with you on the follow-up.

Really appreciate you taking the time — hope to see you back in the `packaging/nix/` corner.

grunisaurus:
Hi @adibhanna and thank you for taking your time to answer my PR. 
I've seen the other PR too, after i published mine :) 

I would love to contribute further on the `/packaging/nix`, will have a loot into that the next days! 

And regarding this: 
> One heads-up if you pick up the desktop build: the `postFixup` here `sed`s the renderer-trust check (`isTrustedRendererUrl`) to allow `nix/store` URLs. That's touching a security-sensitive gate, so I'd want to find a cleaner way to make Electron load from the store than patching that function — happy to work through it with you on the follow-up.
I was also unhappy with this solution, just a quickfix, would love to have a cleaner solution in the future. 

Thanks again for your time and the reply! 
Best regards, Nils

Issue #189, opened 2026-06-15
Any way to sync notes with external storage?


tintinthedev:
This would be SO GOOD for me, really.

I would love to keep my notes somewhere other than my local machine.

I created a [plugin for Logseq](https://github.com/tintinthedev/logseq-sync-plugin) that does exactly that, storing pages in a github repo.

Based on what I searched in ZenNotes docs, this isn't possible (correct me if I'm wrong). So, would you guys like an implementation of it? Storing files on a github repo is enough for me, really.

Maybe it could be a built-in app command.

Tell me what you think. If the answer is positive, I would happily try to implement it!

chaitanya02082000:
yes please , i use obsidian with this feature , would love to switch over, if i can sync in-app to github

tintinthedev:
> yes please , i use obsidian with this feature , would love to switch over, if i can sync in-app to github

I'm working on the implementation right now. I'll probably open a PR later.

sandorex:
I would personally love a generic git based sync so any git capable server is ok (github, gitlab etc)

wedgen:
This is would be useful for the server to provide per file change history, especially if Agents are using the MCP to make changes.

adibhanna:
First-party sync shipped as ZenNotes Cloud in 2.28.0, and the vault is plain `.md` on disk, so any file sync (Syncthing, iCloud, git) works too. The per-file history idea deserves its own issue if you want it; closing this one.

PR #194, opened 2026-06-16
Synchronize notes with remote GitHub repo

tintinthedev:
Adds GitHub synchronization for the desktop build. 
Configure a Github PAT + repo in the UI, then sync notes via pull --rebase → add -A → commit → push.

New command palette entries (Ctrl+Shift+P):
- Sync with GitHub: opens sync modal; auto-starts if PAT + repo already configured
- Configure GitHub Sync: opens config-only modal to change PAT/repo

IPC — 4 new channels (github:get-config, github:set-config, github:list-repos, github:sync) flowing through bridge.ts -> preload -> main process. Web build returns no-ops with an unavailable message.
Config — PAT stored in secret-store (keytar/safeStorage), repo name in PersistedConfig.githubRepo.
Test coverage — 15 vitest cases: listRepos (happy path + API errors), sync (no git, merge conflict, no changes, success, push rejection, connection failure), handleSync (missing PAT/repo).

Code reviews are welcome. Also, this was created based on #189


adibhanna (changes requested):
Thank you for this, @tintinthedev — this is a genuinely useful feature and a strong first cut. The structure is clean (nice separation of `github-sync.ts` from the IPC wiring), you used `execFile` with array args so there's no shell-injection surface, the PAT goes into the OS secret store, the git errors are mapped to friendly user-facing messages, the web build correctly no-ops, and **15 vitest cases** is excellent coverage for a PR like this. Really good instincts throughout.

Because the whole job of this feature is handling a GitHub **credential**, I'm holding it to a high bar on secret-handling. A few things need to change before it can merge — happy to help with any of them.

## 🔴 Security (must-fix)

1. **The PAT is persisted in plaintext on disk.** `setupRemote()` bakes the token into the remote URL (`https://${pat}:x-oauth-basic@github.com/${repo}.git`) and `git remote set-url` writes that to `<vault>/.git/config` in cleartext. That defeats the point of storing it in the secret store — anyone with filesystem/backup access can read it. Please inject the token ephemerally instead, e.g. a `GIT_ASKPASS` helper or `-c http.extraHeader=\"Authorization: Bearer …\"` per invocation, and keep `origin` set to the plain `https://github.com/owner/repo.git`.

2. **The raw PAT is sent to the renderer.** `handleGetConfig()` returns `{ pat, repo }` and that flows over IPC to the renderer (`getGithubConfig()`). Since the renderer displays user-authored Markdown, any renderer-side code execution could call `window.zen.getGithubConfig()` and exfiltrate the token. Please return `{ hasPat: boolean, repo }` and keep the secret main-side only — the config UI only needs to know whether a token is set, not its value.

3. **The PAT can leak into error messages.** `runGit()` rejects with raw `stderr`, and git prints the tokenized remote URL on failure (`fatal: unable to access 'https://TOKEN@github.com/…'`). That string is returned straight to the UI via `GithubSyncResult.error`. Please redact anything matching the token (or the `user:pass@` portion of a URL) before surfacing it.

## 🟠 Correctness

4. **Wrong conflict recovery.** `pullRemote()` runs `pull --rebase --autostash`, but on a conflict it calls `git merge --abort`. Mid-rebase there is no merge to abort, so that command errors and is swallowed by `.catch(() => {})`, leaving the repo **stuck mid-rebase with the autostash unapplied**. It should be `git rebase --abort`.

5. **`git add -A` stages the entire vault.** With no `.gitignore`, this commits and pushes everything in the vault root — `.obsidian/`, attachments, large binaries, anything private the user keeps alongside their notes. Consider scaffolding a sensible `.gitignore` and/or scoping what gets staged, and documenting that the whole vault is pushed.

6. **Hardcoded `main` can hijack existing repos.** `prepareRepo()` does `checkout -B main`, force-moving a vault that's already a git repo onto `main`. Users on `master` or a feature branch would be surprised. Consider detecting/using the existing branch, or making it configurable.

## 🟡 Diff hygiene

- A large part of the `apps/desktop/src/main/index.ts` diff is unrelated Prettier reformatting (the mac window-options indentation, `catch(() => { })`, the deep-link block, etc.) — looks like a local editor/Prettier config that differs from the repo's. Could you revert those so the diff is just the GitHub-sync changes? It'll be much easier to review and won't conflict.

Thanks again — the bones here are good, and items 1–3 are the main thing standing between this and merge. Glad to pair on the credential-injection approach if that's helpful.

tintinthedev:
Thank you for the review! Really good to see that it's really a good improvement. I think I just need some more clarification about certain points you mentioned so I can perform the changes.

About injecting the PAT ephemerally, it would still need to be stored somewhere? If not, then how would the push be made? Could you describe kind of the flow the app would have to take so the PAT can be injected ephemerally? Any representation, like "step 1 -> step 2 -> ..." is already good enough.

Also, about avoiding the staging of the entire vault (point 5), what do you think is a good tactic? Just having a `.gitignore` could still make some unwanted files go to the remote vault. Or, if we were to scope what gets staged, should we perform a recursive search in all directories to find only the `.md` files (probably very slow depending on the amount of directories and sub-directories)?

About choosing the correct branch, I think the best way is to update the feature so it can support an user-selected branch (that can be modified later, just like the repository and PAT). What do you think?

I think that, with these things clear, I can perform the changes. I'm sorry I didn't consider many of the security aspects and conflict recovery.

Also, I'm very sorry about the diff looking messy. It really is something to do with my local neovim setup. I will fix that ASAP and use the project's config.

adibhanna:
Closing: first-party sync shipped as ZenNotes Cloud in 2.28.0, and the plain `.md` vault works with any git workflow outside the app. Thank you for the effort and the clean structure, and sorry I did not get back to your questions in time.

Issue #182, opened 2026-06-15
Quick Task Capture


Bendzae:
First of all, thanks for the great tool. Really enjoying it!

One feature would make it perfect for me:

**Add a streamlined way to capture single tasks via Quick Capture or a separate hotkey.**

I am currently using task tracking apps like Things or TickTick and they allow quick capture of tasks using a global hotkey. I would love to replace them with ZenNotes. However the current quick capture is a bit tedious for this as you have to manually type out  `- [ ]` which creates some friction.

Another solution I could live with is to enable `/` commands in the quick capture pane. As typing `/td` is at least a bit faster.

Maybe you have a better Idea, in the end I just want a quick frictionless way to capture a todo while in a meeting or focused task.

Keep up the great work
Ben

adibhanna:
Done — shipping in **v2.7.0** (8de5e5d). 🎉

Quick Capture now has the same Notion-style `/` slash-command menu as the main editor: type `/` and pick a block (headings, lists, **Task**, quote, code, table, callout, …). Your exact example works — `/td` matches **Task** and inserts `- [ ] `, so capturing a todo is one `/td`+Enter away. Esc closes the menu (without saving/hiding), and the capture window is a bit taller so the full menu shows without clipping.

This implements the `/`-commands approach you suggested (and that @Bendzae sketched in #200). Thanks for the thoughtful request, Ben! 🙏

PR #200, opened 2026-06-18
feat(quick-capture): add `/` autocompletion to the quick capture editor

Bendzae:
## Context
**Disclaimer: mainly ai agent code with manual review, feel free to close if you discourage llm generated code** 

Adds the same Notion-style `/` command autocompletion the main editor has to the Quick Capture editor, which previously had none.

## Changes
- Wire slash-command autocompletion into the Quick Capture editor via the store-free `templateSlashCommandSource` (excludes the store-dependent "Page" command, which has no meaning in the capture renderer). The full menu (headings, lists, todo, quote, code, …) is available.
- Add a high-precedence Esc handler so dismissing an open slash menu only closes the menu, rather than bubbling to the window-level Esc that saves and hides the window — mirroring the existing `completionNavKeymap` pattern.

Tested locally and it seems to work fine.

Related: #182

### DIFF
--- a/packages/app-core/src/components/QuickCaptureApp.tsx
+++ b/packages/app-core/src/components/QuickCaptureApp.tsx
@@ -47,6 +47,15 @@ import { markdownListIndentPlugin } from '../lib/cm-markdown-list-indent'
 import { syntaxHighlighting, HighlightStyle, defaultHighlightStyle } from '@codemirror/language'
 import { tags as t } from '@lezer/highlight'
 import { searchKeymap } from '@codemirror/search'
+import {
+  autocompletion,
+  completionKeymap,
+  completionStatus,
+  closeCompletion
+} from '@codemirror/autocomplete'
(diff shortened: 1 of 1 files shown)


adibhanna:
Thank you for this, @Bendzae! 🙏 I implemented `/` slash commands in Quick Capture directly on the release branch (8de5e5d, shipping in **v2.7.0**), building on your approach:

- Same store-free `templateSlashCommandSource` (full block menu, minus the note-only "Page"), `completionNavKeymap`, `completionKeymap`, and the high-precedence Esc-closes-the-menu handler you added.
- Plus: enlarged the capture window so the menu shows without clipping, and added tests (`cm-slash-commands.test.ts`).

Credited you via `Co-authored-by` on the commit. This also resolves #182. Closing since it's now in `release/v2.7.0` and won't be merged from here — really appreciate the contribution, hope you keep them coming! 🚀

PR #204, opened 2026-06-19
feat: add OpenCode MCP client integration

patricesos:
## Summary

Adds OpenCode as a supported MCP client in Settings → MCP Integrations, matching the existing install/uninstall/status pattern for Claude Code, Claude Desktop, and Codex CLI.

## Changes

- **packages/shared-domain/src/mcp-clients.ts** — adds `opencode` to the `McpClientId` union type and registers the OpenCode descriptor with `serverKey: 'MCP_ZENNOTES'`
- **apps/desktop/src/main/mcp-integrations.ts** — resolves the config path under `~/.config/opencode/opencode.json` ; reads/writes clean JSON for the managed entry; handles install, uninstall, and up-to-date detection

## Verification

Works on **`Windows11 / WSL`**

- [x] Typecheck passes (`npm run typecheck`)
- [x] All 146 tests pass (`npm test`)
- [x] Production build passes (`npm run build`)" 2>&1


adibhanna:
Landed on the **`release/v2.9.0`** branch — thanks @patricesos! 🙏 Shipped as 48e7778 with your authorship preserved.

The integration was well done: it correctly uses OpenCode's own `mcp.<key>` / `type: "local"` config shape (rather than the Claude `mcpServers` format) and matches the existing install / uninstall / status pattern cleanly.

I made one small follow-up on top (e7ebb03): the descriptor hardcoded `serverKey: 'MCP_ZENNOTES'`, so I switched it to the shared `MCP_SERVER_KEY` constant (`'zennotes'`) — that way ZenNotes registers under the same server name in OpenCode as it does for Claude Code/Desktop and Codex, instead of `mcp.MCP_ZENNOTES`. (Also trimmed an unverified vendor attribution from the Settings description.)

Closing since it's merged via cherry-pick rather than the button. Ships in v2.9.0 — thanks for the clean, well-scoped PR! 🙏

Issue #225, opened 2026-06-22
Allow renaming the Tasks view label


farshmaker:
The System Folders settings let you rename Inbox, Quick Notes, Archive, and Trash, but not Tasks. 
Add a "Tasks label" field there so the Tasks view can be customized the same way (sidebar, tab, title bar, command palette).

For example, I'd like to rename it to lowercase names like `todos` or `goals` instead of "Tasks".

<img height="1000" alt="Image" src="https://github.com/user-attachments/assets/8c4676fd-5b5f-4c6f-951d-f0202b9d1e60" />

adibhanna:
On it! good idea!

adibhanna:
Added for the next release (**v2.5.0**), on the `release/v2.5.0` branch.

**Settings → System Folders** now has a **Tasks label** field. Set it to anything — `todos`, `goals`, lowercase, whatever — and it renames the Tasks view everywhere you listed:

- the sidebar row
- the tab title
- the title bar
- the command palette ("Open todos") — still searchable by "tasks"

Clear the field to fall back to "Tasks". It persists like the other labels.

(While there, I also fixed the layout so all the System Folders input boxes are equal width — a long description row was squeezing its box.)

Thanks @farshmaker for the suggestion. Closing as resolved — ships in v2.5.0.

PR #226, opened 2026-06-22
feat(settings): make the Tasks view label customizable

farshmaker:
Add a "Tasks label" field under Settings → Vault → System Folders so the vault-wide Tasks view can be renamed like the inbox, quick, archive, and trash labels. Tasks stays a virtual view, so the label key is modeled as SystemLabelKey = NoteFolder | 'tasks' rather than touching NoteFolder.

The resolved label is used everywhere the view surfaces: sidebar row, tab title, title bar, buffer palette, command palette, and the view heading.

Related: #225 

screenshot 2026-06-22 at 14 42 08
screenshot 2026-06-22 at 14 42 56

Palette:
screenshot 2026-06-22 at 14 43 50
Buffer:
screenshot 2026-06-22 at 14 54 12

### DIFF
--- a/packages/app-core/src/components/BufferPalette.tsx
+++ b/packages/app-core/src/components/BufferPalette.tsx
@@ -83,7 +83,7 @@ function buildEntries(deps: BuildDeps): BufferEntry[] {
     if (isTasksTabPath(path)) {
       entries.push({
         path,
-        title: 'Tasks',
+        title: labels.tasks,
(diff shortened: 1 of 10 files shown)


adibhanna:
Shipped this! will close PR

farshmaker:
> Shipped this! will close PR

it looks like that "Open buffers" pop up still show the default name "Task" instead of the one set in the settings. My PR took this into account

PR #274, opened 2026-06-26
chore: fix attachments typo

jcdiv47:
Chore: fixes the typo `attachements`(which should be `attachments`) across the codebase.

_Disclaimer: this PR is driven by Codex but I have manually checked every instance to make sure we are changing exactly the places needed._

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ ZenNotes now works better with existing Obsidian-style vaults.
 - loose files anywhere in the vault are surfaced as files/assets
 - embedded files like `![[image.png]]` resolve more like Obsidian
 - new referenced files default to the vault root instead of a required attachments folder
-- legacy `attachements/` and `_assets/` folders are still recognized
+- legacy `attachments/` and `_assets/` folders are still recognized
 
 This means imported vaults with top-level notes, folders, and loose images/files behave much more naturally.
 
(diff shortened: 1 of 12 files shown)


adibhanna:
Thanks for this — and especially for manually auditing every instance! Unfortunately we can't take this one, because the misspelled `attachements` is **intentional**.

It's the literal on-disk folder name ZenNotes used to save attachments into back in the 1.x line, so users who added attachments in those versions have real `attachements/` folders on disk (with notes referencing `attachements/…`). Since the assets-unification work (#185), the code keeps that exact (ugly) spelling on purpose — to **recognize and migrate** those legacy folders into `assets/`. The desktop app, the MCP server, and the Go server all have to agree on that string.

Renaming the constant value `'attachements'` → `'attachments'` would make the migration look for a folder that isn't there, so it'd silently skip those legacy folders and break embeds for anyone upgrading from an older vault. (Note the *identifiers* like `PRIMARY_ATTACHMENTS_DIR` are already spelled correctly — it's only the string value that has to stay "wrong," which is exactly the kind of thing that reads as a typo but isn't.)

Really appreciate the contribution, and the careful write-up. 🙏

As a follow-up on our side: we may *add* `attachments/` (correct spelling) as an **additional** recognized folder alongside `attachements`/`_assets`, so correctly-spelled vaults are picked up too — without dropping the legacy name.

jcdiv47:
Thank you for the detailed explanation! 🙏 I should have given it a more thorough thought and/or opened an issue before raising a PR and asking for your attention/time. Been enjoying the app and I thank you for it.

Issue #273, opened 2026-06-26
Space as leader does not work on 2.8.0 on Linux


littlelinuxoid:
After the 2.8.0 release, the app stopped recognizing Space as a leader key. Space presses in normal mode seems to just get ignored. Remapping leader to other keys like Tab or Enter works as expected, but my muscle memory is screaming at this.

adibhanna:
Ill fix this asap! probably tomorrow

madhu-basavanna:
its working fine for me on debian 13 with .deb package. Just wanted to give info if it helps for the developer

theurzil:
@littlelinuxoid  I created a PR for it, could u test it and see if solves the issue for you !

littlelinuxoid:
@theurzil i'm sorry, i am not very familiar with electron development and this codebase in particular, and i wasn't able to build and run the project. 
After cloning the repo, checking out your commit, following the guide.md file:
`npm ci`
`npm run dev:desktop`

produces 
`error during start dev server and electron app:
Error: Electron uninstall
    at getElectronPath (file:///home/ivan/Programming/probe/zennotes/node_modules/electron-vite/dist/chunks/lib-BmEkZIgk.mjs:129:19)
    at startElectron (file:///home/ivan/Programming/probe/zennotes/node_modules/electron-vite/dist/chunks/lib-BmEkZIgk.mjs:198:26)
    at createServer (file:///home/ivan/Programming/probe/zennotes/node_modules/electron-vite/dist/chunks/lib-t2ExBjL5.mjs:74:14)
    at async CAC.<anonymous> (file:///home/ivan/Programming/probe/zennotes/node_modules/electron-vite/dist/cli.mjs:67:9)`

what do i do?

theurzil:
@littlelinuxoid  could you run `npm rebuild electron esbuild` instead of `npm ci`

littlelinuxoid:
@theurzil `npm rebuild electron esbuild` instead of `npm ci` and `npm run dev:desktop` afterwards produces `sh: line 1: electron-vite: command not found`

theurzil:
@littlelinuxoid  could try it with this order then: 
npm ci
npm rebuild electron esbuild
npm run dev:desktop

littlelinuxoid:
@theurzil okay, sorry for all that, i just redirected npm to my system-wide electron installation which is working as expected. 

[...]

PR #280, opened 2026-06-26
fix(vim): let Space leader fire on the home view (#273)

theurzil:
This is the for the issue #273
The home view's keydown bail swallowed every key while it was focused, including the leader. Yield to it for its own nav keys but let the leader and pending leader sequences fall through to VimNav.

### DIFF
--- a/packages/app-core/src/components/VimNav.tsx
+++ b/packages/app-core/src/components/VimNav.tsx
@@ -9,7 +9,8 @@ import {
   isEditorInsertMode,
   isEditorFocused,
   isVimAwaitingArgument,
-  resolveNextPanel
+  resolveNextPanel,
+  shouldYieldToHomeNav
 } from '../lib/vim-nav'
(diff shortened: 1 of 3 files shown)


adibhanna:
Landed on the **`release/v2.9.0`** branch 🎉 — shipped as 884d79673afc9b14b2611fd55a83e84d40a8903a with your authorship preserved.

I cherry-picked it onto the release branch rather than merging directly, because it had a one-line conflict in `vim-nav.test.ts` — purely an import collision with some concurrent calendar-nav work on that branch (`getVisiblePanels`/`resolveNextPanel` vs your `shouldYieldToHomeNav`), nothing in the fix itself. Resolved that, and verified `typecheck` + the full suite green, including all four of your `shouldYieldToHomeNav` cases.

Clean fix with solid test coverage — exactly the kind of PR that's a pleasure to merge. Ships in v2.9.0. Thanks! 🙏

PR #308, opened 2026-07-02
ci: fix Linux arm64 release by using a native fpm

### LINKED ISSUE #307: Linux arm64 release job fails: bundled fpm is x86-only (deb/pacman "Exec format error")
Well that's embarrassing...

PR #270 added a native ubuntu-24.04-arm leg to the release matrix, but I validated it in isolation (a CI-style build) instead of against a real release run that exercises the full packaging ecosystem. The gap: the failure only surfaces in the deb/pacman packaging step, which my isolated check never reached. So v2.9.0 shipped without Linux arm64 assets. That's on me.

## What's happening

It looks like electron-builder only bundles an x86 fpm binary (fpm-1.9.3-2.3.1-linux-x86). The deb and pacman targets shell out to it. On the native aarch64 runner that x86 binary errors:

```
.../fpm-1.9.3-2.3.1-linux-x86/lib/ruby/bin.real/ruby: cannot execute binary file: Exec format error
```

The AppImage and tar.gz targets don't use fpm, so they build fine first; the job then dies at the deb step, failing the whole dist:linux run. Net result: no arm64 artifacts uploaded, while every other platform succeeds.

Evidence:
- Failed run: Release Linux arm64 job in run [28545310913](https://github.com/ZenNotes/zennotes/actions/runs/28545310913)
- v2.9.0 assets include mac-arm64 but no linux-arm64 .deb/.pacman/.AppImage/.tar.gz

## Fix (incoming PR)

Install a native fpm on the arm64 leg and set `USE_SYSTEM_FPM=true` so electron-builder uses it; x64 keeps the bundled binary (it works there). I'm verifying the fix on my fork — first an isolated arm64 build, then a real `v2.9.1` fork release with the arm64 assets actually attached, which I'll link here as proof before opening a new PR.

My apologies.

robwilkerson:
Fixes #307

## Problem

electron-builder bundles only an x86 `fpm` binary, which the `deb` and `pacman` targets shell out to. On the native `ubuntu-24.04-arm` runner it can't execute:

```
cannot execute binary file: Exec format error
```

`AppImage` and `tar.gz` don't use `fpm`, so they build first and the job only dies at the `deb` step, failing the whole `dist:linux` run. The result: v2.9.0 shipped with no Linux arm64 assets while every other platform succeeded.

## Fix

On the arm64 leg only, install a native `fpm` and set `USE_SYSTEM_FPM=true` so electron-builder packages `deb`/`pacman` with it. The x64 leg is unchanged and keeps using the bundled binary.

## Verification

Full release matrix run on my fork. Linux x64, Linux arm64, and Windows all built and uploaded; macOS fails only because my fork lacks the code-signing secrets, which is unrelated to this change.

- Proof release, arm64 `.deb`/`.pacman`/`.AppImage`/`.tar.gz` attached: https://github.com/robwilkerson/zennotes/releases/tag/v2.9.2
- Green Linux arm64 job: https://github.com/robwilkerson/zennotes/actions/runs/28559119900/job/84672925443
- Full run: https://github.com/robwilkerson/zennotes/actions/runs/28559119900

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -94,7 +94,21 @@ jobs:
         # `bsdtar` (libarchive-tools) is required by electron-builder's pacman
         # target to generate the package .MTREE; it isn't on the GitHub Ubuntu
         # runners (x64 or arm64).
-        run: sudo apt-get update && sudo apt-get install -y libarchive-tools
+        #
+        # electron-builder only bundles an x86 `fpm`, so the deb/pacman targets
+        # fail with "Exec format error" on the native arm64 runner. Install a
+        # native fpm there and switch electron-builder to it via USE_SYSTEM_FPM.
+        # Scope the flag to arm64 by exporting it to $GITHUB_ENV only here:
+        # defining USE_SYSTEM_FPM at all (even empty) makes electron-builder use
+        # a system fpm, which the x64 runner doesn't have.
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libarchive-tools
+          if [ "${{ matrix.name }}" = "Linux arm64" ]; then
+            sudo apt-get install -y ruby ruby-dev build-essential
+            sudo gem install fpm --no-document
+            echo "USE_SYSTEM_FPM=true" >> "$GITHUB_ENV"
+          fi
 
       - name: Build release artifacts
         run: ${{ matrix.command }}


adibhanna (approved):
Approving. 🚀 Clean, correctly-scoped fix — keeping `USE_SYSTEM_FPM` inside the arm64 `if` (exported via `$GITHUB_ENV`) so x64 stays on the working bundled binary is exactly the right call; that subtlety is easy to get wrong.

And thank you for the verification: a real fork release (`v2.9.2`) with the arm64 `.deb`/`.pacman`/`.AppImage`/`.tar.gz` actually attached is precisely what closes the gap that let this slip through — so, genuinely, no apology needed at all. Great catch, precise diagnosis, textbook fix. Merging now; it ships with v2.10.0. 🙏

PR #325, opened 2026-07-03
fix(tables): preserve latex backslashes and escape pipes (#319)

### LINKED ISSUE #319: KaTeX not rendering in preview inside of tables
Image

Inside of the preview mode, while inside of tables, KaTeX does not render correctly. This is frustrating because while in "edit" mode, it renders it perfectly, although it is quite frustrating to write the KaTeX inside of the tables. 

Edit mode version: 

Image

heyitsadityaa:
Preview KaTeX inside tables was broken after editing because table serialization turned `\sum` into `\\sum` and `$P \longrightarrow Q$` into `P \\longrightarrow Q`. This change preserves LaTeX backslashes while still escaping literal pipes, so Preview renders math correctly again. Closes #319.

### DIFF
--- a/packages/app-core/src/lib/markdown-table.test.ts
+++ b/packages/app-core/src/lib/markdown-table.test.ts
@@ -82,6 +82,27 @@ describe('parseTable', () => {
 })
 
 describe('serializeTable round-trip', () => {
+  it('preserves latex backslashes on serialize', () => {
+    const t = setCell(parse(SIMPLE), { row: 0, col: 0 }, '\\sum')
+    const out = serializeTable(t)
+    expect(out).toContain('\\sum')
+    expect(parse(out).rows[0][0]).toBe('\\sum')
+  })
+
+  it('still escapes literal pipes', () => {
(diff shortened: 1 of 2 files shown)


adibhanna:
Thanks @heyitsadityaa! I brought this into the `release/v2.11.0` branch, so it ships with 2.11.0.

I verified it resolves the reported `\sum` case. While reviewing I found one round-trip gap: `escapeCell` stopped escaping backslashes, but `unescapeCell` still un-escaped `\\`, so a cell holding a LaTeX line break `\\` (matrices, aligned math) collapsed to `\` when parsed back. I added a one-line follow-up making `unescapeCell` symmetric (only un-escape `\|`) plus a double-backslash regression test, layered on top of your commit.

It will land on `main` when 2.11.0 merges. Appreciate the fix and the tests 🙏

adibhanna:
Closing — this shipped in `release/v2.11.0` earlier (merged as `5909801` / `f1a90e6`, plus follow-ups keeping unescape symmetric and rendering raw pipes in math). It'll be in 2.11.0. Thanks again @heyitsadityaa.

PR #331, opened 2026-07-04
feat(vim): add ':sav' and ':saveas' command to save active note with …

### LINKED ISSUE #265: [feat] :w / :sav / :saveas implementation to save file
:w/:save/:saveas to save the file and auto update the file title in the top pane

Image

heyitsadityaa:
Implements Vim-style save commands for the editor:
- `:sav` / `:saveas` saves the active note under a new title

The save-as flow persists the current note state first, then renames it via the existing `renameNote` path. The backend handles file extension and folder placement. Pane tabs update automatically through `renameNoteState`.

Closes #265

### DIFF
--- a/packages/app-core/src/components/Editor.tsx
+++ b/packages/app-core/src/components/Editor.tsx
@@ -362,6 +362,11 @@ function registerVimCommands(): void {
   Vim.defineEx('write', 'w', () => {
     void useStore.getState().persistActive()
   })
+  Vim.defineEx('saveas', 'sav', (_cm: unknown, params: { argString?: string } | undefined) => {
+  const newName = (params?.argString ?? '').trim()
+  if (!newName) return
+  void useStore.getState().saveActiveNoteAs(newName)
+})
   Vim.defineEx('format', 'format', () => {
     void useStore.getState().formatActiveNote()
   })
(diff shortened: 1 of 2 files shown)


adibhanna:
Thanks @heyitsadityaa. Two things before this can land:

1. **Save-as semantics.** This *renames* the note (the original is gone), but Vim's `:saveas` writes a copy under the new name and keeps the original file. So `:saveas backup` would surprise someone expecting a copy — a potential data-loss footgun. Either make it a true copy-and-switch (matching Vim), or, if a rename is what you want, name the command `:rename`/`:file` so it's not mistaken for Vim's copy behavior.
2. **Formatting.** The `defineEx` body is indented 2 spaces where the surrounding code uses 4 — it'll fail the formatter.

Happy to help land it once the save-as semantics are sorted out.

adibhanna:
Merged into `release/v2.11.0` (6601bd2), shipping in 2.11.0. Thanks @heyitsadityaa!

I reworked the implementation to match Vim's actual `:saveas` semantics: it now writes a **copy** under the new name and **keeps the original**, rather than renaming the note (which would lose the original). It saves the current note, duplicates it, renames the copy to the requested name, and opens it. Also fixed the `defineEx` body indentation.

Verified in-app over CDP: after `:saveas Foo`, both the original and the new `Foo.md` exist on disk with the content, and the editor switches to `Foo`.

Issue #151, opened 2026-06-13
Hint mode broken in task view


DaveDavenport:
2.3.0 moves hint mode after the leader command. (`leader` `h`)

Now in the task view, the default leader command (space) marks the task as done, stopping the user from using hint mode. This breaks my keyboard driven view, and forcing me to move to mouse (and accidentally marking items as done).

DaveDavenport:
Even if I move leader, I am unable to enter hint mode.

adibhanna:
Thanks @DaveDavenport — and you were right that moving the leader didn't help.

**Root cause:** when hint mode moved to `<leader>h` in 2.3.0, VimNav was left bailing out of the Tasks/Tags views for every key *except* the old standalone `f` hint key. That killed the leader itself in those views — so hint mode (and every other leader command) was unreachable there, no matter which key the leader was bound to. On top of that, the Tasks list binds Space to toggle-done, which collided with the default Space leader.

**Fix (`c30eaf8`, shipping in 2.4.0):**
- The leader now works in the Tasks and Tags views. `<leader>h` enters hint mode (and `<leader>o`, etc. all work). VimNav arms and consumes the leader before the panel sees it, so with the default leader, **toggle-done moves to `x`** (Space still toggles if you've remapped your leader off Space — so your "move leader" workflow now gives you both).
- While the hint overlay is open, the Tasks/Tags handlers yield every key to it, so hint labels that overlap panel keys (`j`/`k`/`x`/`o`) are no longer stolen.
- Bonus from testing: **Esc no longer closes the Tasks/Tags tab** (they're tabs like any note tab). Esc clears an active `/` filter; close with `:q`, the header ✕, or ⌘W.

Closing as fixed — please reopen if anything in the Tasks view still doesn't take the keyboard.

PR #333, opened 2026-07-05
fix: Esc closes Tasks view (#322)

### LINKED ISSUE #322: Tasks View: Esc does not close the view despite tooltip indicating it should
- **Date:** 2026-07-03
- **Severity:** Minor
- **Environment:**
  - **OS:** Linux
  - **DE:** Niri
  - **Compositor:** Wayland
- **Software version:** 2.10.0

## Description

Hovering over the **Close** button in Tasks View displays the tooltip:

> Close (:q or Esc)

However, pressing `Esc` does not close the view.

## Steps to reproduce

1. Open Tasks View.
2. Hover over the **Close** button and observe the tooltip.
3. Press `Esc`.
4. Press `:q` for comparison.

## Expected Behaviour

Both `Esc` and `:q` should close the Tasks View, as indicated by the tooltip.

## Actual Behaviour

- `:q` closes Tasks View.
- `Esc` has no effect.

## Frequency

Always.

## Notes

This is either an incorrect tooltip or `Esc` is not correctly bound to the close action in Tasks View. Either the tooltip should be updated or the `Esc` shortcut should be implemented to match the documented behaviour.

heyitsadityaa:
The Close button tooltip advertised "Close (:q or Esc)" but pressing Esc
only cleared the active filter, it never closed the view. This made the
tooltip misleading.

Fix: first Esc clears the filter (if any), second Esc closes the view.
This matches the common "Esc to retreat" UX pattern and aligns behavior
with the existing tooltip.

Fixes #322.

### DIFF
--- a/packages/app-core/src/components/TasksView.tsx
+++ b/packages/app-core/src/components/TasksView.tsx
@@ -304,11 +304,13 @@ export function TasksView(): JSX.Element {
       }
 
       if (key === 'Escape') {
-        // Tasks is a tab like a note tab — Esc clears an active filter but must
-        // never close the tab (other tabs don't close on Esc). Close with :q,
-        // the header ✕, or ⌘W. (#151)
+        // First Esc clears filter, second Esc closes view.(#322)
         consume()
-        if (filter) setFilter('')
+        if (filter) {
+          setFilter('')
+        } else {
+          closeTasksView()
+        }
         return
       }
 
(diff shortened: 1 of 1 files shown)


adibhanna:
Thanks for tackling this @heyitsadityaa, and good eye on the tooltip mismatch. #322 has since been resolved a different way and closed: we kept `Esc` as clear-filter-only (never closing the tab), to stay consistent with note tabs and the deliberate #151 behavior, and fixed the misleading tooltip instead — it now reads `Close (:q)` (shipping in 2.11.0). Making `Esc` close the view would reverse that #151 decision and leave the corrected tooltip inconsistent again, so I'm going to close this PR. The tooltip-vs-behavior mismatch you spotted is exactly what got fixed, though — appreciate it.

Issue #278, opened 2026-06-26
Excalidraw creation command via Command Palette


junereycasuga:
Currently, to create a new Excalidraw drawing, a user must manually use the mouse to click the `+` icon and select "New drawing" from the UI. For a Vim-focused, keyboard-first note-taking app, having to break hand placement on the home row to reach for a mouse disrupts flow, especially when other main features (like creating a new note or a new database) are fully accessible via keyboard shortcuts or the command palette.

I would like an action added to the Command Palette (`Cmd+Shift+P` / `Ctrl+Shift+P`) to quickly create a new drawing. 

* **Command Name Suggestion:** `New Drawing`
* **Behavior:** Triggering this command should create a new canvas tab immediately, bypassing the need to use the sidebar or mouse dropdown menus.

In addition, it would also be nice to have the following commands in the palette:
- `Embed a drawing` - embeds an existing Excalidraw drawing into the note
- `Embed a new drawing` - opens a new Excalidraw tab and automatically embeds the drawing into the note

Another excalidraw embedding request is described in https://github.com/ZenNotes/zennotes/issues/276

adibhanna:
Added in v2.13.0 (on `release/v2.13.0` in 1c5493a). The command palette now has **New Drawing** (create + open a `.excalidraw`), **Embed Existing Drawing…** (searchable picker that inserts an `![[…]]` embed), and **Embed New Drawing** (create, insert an embed at the cursor, then open it). Implemented by @junereycasuga in #360. Thanks!

PR #360, opened 2026-07-10
app-core: add excalidraw drawing embeds

### LINKED ISSUE #276: Automatically resolve and render/link converted Obsidian Excalidraw embeds
**Context**
---
In ZenNotes v2.8.0, support was added to import and convert Obsidian Excalidraw drawings into native ZenNotes drawings (#266). This works great for standalone drawing files. However, when migrating full markdown notes that contain embedded drawing attachments, the Obsidian wiki-link syntax for the embed is broken.

**Current Behavior**
---
When a markdown note contains an inline embed like ![[Observability.excalidraw]], ZenNotes renders it as plain, unresolvable text in the editor/preview. The user has to manually find the converted ZenNotes drawing and re-link or re-embed it.

Image

Compared to how it would look like in Obsidian

Image

**Proposed Solution / Enhancements**
---
It would be ideal if ZenNotes could parse these internal link attachments and check if a native ZenNotes drawing counterpart exists.

    Option A (Preferred): Dynamic Rendering If Observability.excalidraw has already been converted to a native ZenNotes drawing, automatically render the native interactive drawing right inside the note where the ![[...] tag is placed.

    Option B (Fallback): Native Link Replacement
    If rendering directly inline is too heavy, automatically transform the broken text into a clean internal link pointing directly to the converted ZenNotes Excalidraw file instead of the old Obsidian path.

**Why this is helpful**
---
This would completely bridge the gap for users migrating large Obsidian vaults to ZenNotes, ensuring that existing notes with embedded diagrams don't suffer from broken references after the drawings are converted.

junereycasuga:
### Summary

Adds first-class Excalidraw drawing support in notes.

Drawings can now be created, opened, embedded, and previewed inline from markdown using the same wikilink embed syntax as images:

```
![[diagram.excalidraw]]
![[diagram.excalidraw|600]]
![[diagram.excalidraw|600x400]]
```

Would close https://github.com/ZenNotes/zennotes/issues/276 and https://github.com/ZenNotes/zennotes/issues/278

### Changes

- Add **New Drawing** command to create and open a .excalidraw file.
- Add **Embed Existing Drawing…** command with a searchable drawing picker.
- Add **Embed New Drawing** command to create a drawing, insert an embed at the cursor, then open the drawing.
- Render `.excalidraw` and Obsidian `.excalidraw.md` embeds as PNG previews in:
- editor live preview
- rendered preview mode
- Cache rendered previews in-memory by path + mtime.
- Invalidate previews when the drawing file changes.
- Click embedded previews to open the drawing in an Excalidraw tab.
- Prevent note transclusion from expanding .excalidraw files as raw JSON.
- Update Help docs for creating and embedding drawings.

### Demo

https://github.com/user-attachments/assets/c706da30-f7a2-495b-94d9-3a263eb3480e

### Testing

- Added tests for:
	- Excalidraw embed size hints
	- Excalidraw embed path resolution
	- Markdown rendering of Excalidraw embed placeholders

### DIFF
--- a/packages/app-core/src/App.tsx
+++ b/packages/app-core/src/App.tsx
@@ -170,6 +170,11 @@ const TemplatePalette = lazy(async () => {
   return { default: module.TemplatePalette }
 })
 
+const EmbedDrawingPalette = lazy(async () => {
+  const module = await import('./components/EmbedDrawingPalette')
+  return { default: module.EmbedDrawingPalette }
+})
+
 const SettingsModal = lazy(async () => {
   const module = await import('./components/SettingsModal')
   return { default: module.SettingsModal }
@@ -256,6 +261,8 @@ function App(): JSX.Element {
   const setOutlinePaletteOpen = useStore((s) => s.setOutlinePaletteOpen)
   const templatePaletteOpen = useStore((s) => s.templatePaletteOpen)
   const setTemplatePaletteOpen = useStore((s) => s.setTemplatePaletteOpen)
+  const embedDrawingPaletteOpen = useStore((s) => s.embedDrawingPaletteOpen)
+  const setEmbedDrawingPaletteOpen = useStore((s) => s.setEmbedDrawingPaletteOpen)
   const sidebarOpen = useStore((s) => s.sidebarOpen)
   const noteListOpen = useStore((s) => s.noteListOpen)
   const zenMode = useStore((s) => s.zenMode)
@@ -685,6 +692,11 @@ function App(): JSX.Element {
         focusEditorNormalMode()
         return
       }
+      if (e.key === 'Escape' && state.embedDrawingPaletteOpen) {
+        setEmbedDrawingPaletteOpen(false)
+        focusEditorNormalMode()
+        return
+      }
       if (e.key === 'Escape' && state.outlinePaletteOpen) {
         setOutlinePaletteOpen(false)
         focusEditorNormalMode()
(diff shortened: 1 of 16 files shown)


adibhanna:
Thanks @junereycasuga, this is a really nice, well-tested feature and it slotted into the existing embed pipeline cleanly. Merged into `release/v2.13.0` in 1c5493a (ships in v2.13.0), credited to you as co-author.

I gave it a thorough review (no regressions to `![[image]]`/`![[Note]]` embeds, no XSS/traversal, no blob leak, and the heavy Excalidraw dep is genuinely lazy-loaded), and polished two things on the way in:

- **Bounded the preview cache.** `previewCache` had no cap, and each entry is a full @2x base64 PNG, so browsing a vault with many drawings would slowly grow the heap. Added a 24-entry LRU (mirroring the markdown cache) and removed `invalidateAllExcalidrawPreviews`, which had no callers.
- **Fixed a reading-view re-mount flash.** The preview pane rebuilds its DOM on every render, tearing down each embed's React root, so even a cache hit flashed the loading pulse. `ExcalidrawPreview` now seeds from a synchronous cache peek and holds the last image while it re-resolves.

Verified end to end in-app: an embedded drawing renders as an inline PNG at full size and with a `|240` size hint, no stuck loaders. Your 28 tests pass.

A few non-blocking notes for later: bare `![[Drawing]]` (no extension) still renders as a link rather than a preview; the picker's secondary label shows the folder enum rather than a subfolder path; and there's no dedicated keybinding for the three commands yet. None block the merge.

Closing since it has landed on the release branch (it reaches `main` when v2.13.0 ships). Thanks again for the excellent contribution!

PR #361, opened 2026-07-10
feat: add toast notification on PDF export

### LINKED ISSUE #257: [feat] Add a notification about export PDF
## Feature Description
After successfully exporting a note to PDF, there is currently no feedback to the user that the export has completed. Adding a success notification would improve the user experience by confirming the action finished and, ideally, providing quick access to the exported file.

## Motivation
Without any confirmation, users are left uncertain whether the export completed successfully or silently failed, especially on longer notes where the export may take a moment. A brief notification closes this feedback loop.

## Proposed Behavior
After a successful PDF export, display a toast/snackbar notification that:
- Confirms the export was successful (e.g. *"PDF exported successfully"*)
- Optionally includes a quick action to open or locate the saved file (e.g. *"Open file"* / *"Show in folder"*)
- Dismisses automatically after a few seconds

On failure, the notification should indicate that the export failed and suggest a next step.

mrsaynothing:
## Summary

Add a toast notification system and show a success toast after exporting a note to PDF. On failure, show an error toast alongside the existing `window.alert()`. On desktop, the success toast includes a "Show in folder" action button that reveals the saved PDF in the OS file manager.

## Why

Previously, users had no feedback after a PDF export completed successfully, leaving them unsure whether the export worked or silently failed. This closes that feedback loop.

## Testing

1. Open a note and press the export PDF shortcut/command
2. On success: a toast reading "PDF exported successfully" appears in the bottom-right corner and auto-dismisses after 4 seconds. On desktop, clicking "Show in folder" opens the saved PDF location
3. On failure (e.g. unsaved changes): an error toast appears alongside the existing `window.alert()` with the error details
4. Verify typecheck passes: `npm run typecheck --workspace @zennotes/app-core`

Closes #257


adibhanna:
Thanks @wholetthekfcgo, nice feature and a genuinely useful one. Merged into `release/v2.13.0` in bf00608 (ships in v2.13.0), credited to you as co-author.

I reworked a few things on the way in:

- **Web regression:** the success path closed `preparedExportWindow`, but on the web runtime that window *is* the print/PDF view (`exportNotePdf` navigates the same named window). I dropped that close and instead scoped the success toast to `if (pdfPath)`, since a returned path means a real file was written on disk (desktop). Web returns `null` (the print window is the feedback there) and desktop returns `null` on a cancelled save dialog, so no misleading "exported" toast in those cases.
- **Design system:** reworked `Toast.tsx` to use the `Button`/`IconButton` primitives, the SVG icon set, and the `success`/`danger` theme tokens (styled to match the existing `AppUpdateNotice`) instead of raw `<button>`s and emoji.
- **Error UX:** dropped the redundant `window.alert()` — the error toast replaces it, and error toasts now stay until dismissed (they don't auto-dismiss like success/info) so they're not missed.
- Added tests for the toast store + host.

Your `revealFilePath` bridge method and the overall approach carried straight through. Closing this since it has landed on the release branch (it reaches `main` when v2.13.0 ships). Thanks again!

PR #364, opened 2026-07-10
fix: follow app theme in Excalidraw (fixes #363)

### LINKED ISSUE #363: can't set dark mode for Excalidraw
Image

I can't find the typical toggle that Excalidraw provides. 
Since ZenNotes hijacks the keyboard, I can't use Command Pallete nor dedicated Excalidraw keybind for toggling the theme
Most likely you could pass the ZenNotes's `currentTheme` as a prop to Excalidraw component to use either dark or light theme I think

The issue is with [my custom kanagawa-paper theme](https://github.com/Kamyil/.dotfiles/blob/main/zennotes/themes/kanagawa-paper/theme.css). I mean, not the theme itself but rather the Excalidraw usage, since it looks in it's built-in themes registry. I can propose the PR that could make Excalidraw support detecting dark/light from custom color schemes

Kamyil:
fixes #363
[My custom theme](https://github.com/Kamyil/.dotfiles/blob/main/zennotes/themes/kanagawa-paper/theme.css) already declares dark mode in the [manifest](https://github.com/Kamyil/.dotfiles/blob/c7af8262a71c7d57bd381a70666b122d82b0e79a/zennotes/themes/kanagawa-paper/manifest.json#L6):

```json
  {
    "modes": "dark"
  }
```

The problem is that the old Excalidraw code checked only the built-in THEMES registry:

```ts
  THEMES.find((t) => t.id === themeId)
```

My custom theme ID:

```text
  custom-kanagawa-paper
```

is not in that registry, so the lookup returned undefined regardless of the manifest.

The fix needs to be in ZenNotes code: Excalidraw must use ZenNotes’ resolved mode:

```ts
  document.documentElement.dataset.themeMode
```

co-oped with GPT 5.6 Luna (low) agent

### DIFF
--- a/packages/app-core/src/components/ExcalidrawView.test.ts
+++ b/packages/app-core/src/components/ExcalidrawView.test.ts
@@ -1,0 +1,110 @@
+// @vitest-environment jsdom
+//
+// Guards: ExcalidrawView derives its theme from document.documentElement.dataset.themeMode
+// rather than looking up the theme id in THEMES. The THEMES array only contains built-in
+// themes, so a custom theme id (e.g. "custom-mine") would always resolve to "light" under
+// the old approach — THEMES.find returns undefined and undefined?.mode === 'dark' is false.
+// This test verifies that dark custom themes correctly produce a dark Excalidraw theme,
+// and would catch a regression back to THEMES.find().
+
+import { act, createElement } from 'react'
+import { createRoot, type Root } from 'react-dom/client'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+// Spy on the Excalidraw component so we can inspect its theme prop.
+// vi.hoisted is needed so the variable is in scope for both the mock module factory
+// and the test body.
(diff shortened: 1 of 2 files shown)


adibhanna:
Thanks @Kamyil, great diagnosis and a proper regression test to go with it. Merged into `release/v2.13.0` in 85afc2b (ships in v2.13.0), credited to you as co-author.

Your core fix was spot on: custom theme ids aren't in the built-in `THEMES` registry, so deriving the mode from `data-theme-mode` is the right call. One tweak on the way in: `data-theme-mode` is written in an App `useEffect` (after render), so reading it during render (via the two throwaway `useStore` subscriptions) could leave an open drawing a frame stale on a live theme switch, and OS/auto dark-mode changes wouldn't refresh it at all. I swapped that for a small `MutationObserver` on `<html data-theme-mode>` so an open drawing tracks live theme and system changes too, and added a test covering that. Your two tests were kept as-is.

Closing this since it has landed on the release branch (it reaches `main` when v2.13.0 ships). Thanks again for the contribution!

PR #365, opened 2026-07-10
feature: custom Kanagawa Papek Ink theme variant

Kamyil:
It's a port of best Kanagawa variant theme for neovim (with muted colors:
https://github.com/thesimonho/kanagawa-paper.nvim

but with custom, little darker background for better focus

image
image
image

color pallette: https://github.com/thesimonho/kanagawa-paper.nvim/blob/master/palette_colors.md

### DIFF
--- a/packages/app-core/src/lib/help.ts
+++ b/packages/app-core/src/lib/help.ts
@@ -743,7 +743,7 @@ export const HELP_SETTINGS: HelpSettingsSection[] = [
   {
     title: 'Appearance',
     items: [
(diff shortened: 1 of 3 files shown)


adibhanna:
Thanks @Kamyil, lovely variant, the muted palette on the dark canvas looks great. Merged into `release/v2.13.0` in c6cf45a (ships in v2.13.0), credited to you as co-author.

One fix on the way in: the diff changed six `:root` selectors in the shared glass-alpha rule to `::root`. `::root` isn't valid CSS, and a single invalid selector in a comma-separated list drops the whole rule, which would have knocked out the glass variables for every dark theme in that list (one-dark, nord, tokyo-night, kanagawa, rosé-pine, black-metal). I restored `:root` and added `kanagawa-paper-ink` to that list correctly. The theme registration, its token block, and the help entry all carried through as-is.

Verified in-app: `data-theme` resolves to `kanagawa-paper-ink` and the palette renders as expected. Closing since it has landed on the release branch (it reaches `main` when v2.13.0 ships). Thanks again!

PR #385, opened 2026-07-13
feat: preserve PDF URL hash fragments (#page, #search) in links and embeds

### LINKED ISSUE #295: [feat] Extended PDF Integration
It would dramatically better the workflow if you could open the linked PDF in the Browser to use its already existing Features (like Search or Queries like jumping to pages) or extend the embedded view with search and queries directly 

example:

[Script](script.pdf#page=12&search=<Keyword>) -> Jumps to page 12 + searches for <keyword> so it is focused

( usecase: link notes directly to the content of the pdf so you dont have to copy the context of the PDF into the note directly)

mrsaynothing:
## Problem

Markdown links to PDFs with URL hash fragments (e.g. `#page=3`, `#search=keyword`) lose the fragment at multiple points in the resolution pipeline. Chromium's built-in PDF viewer natively supports these fragments, but they never reach it.

Closes #295

## Before

`[Guide](guide.pdf#page=3&search=installation)` → PDF opens at **page 1** (fragment stripped)

## After

`[Guide](guide.pdf#page=3&search=installation)` → PDF opens at **page 3** with "installation" highlighted

Works everywhere: inline embeds, the pinned reference pane, preview anchors, and Cmd/Ctrl+click in the editor.

## Example

```markdown
[Configuration](guide.pdf#page=3)
[Search: installation](guide.pdf#search=installation)
[Troubleshooting + search](guide.pdf#page=5&search=installation)
![[guide.pdf#page=3&search=installation]]
```

## Supported Chromium PDF parameters

Since we pass the fragment through to the native viewer, all standard parameters work:
- `#page=N` — jump to page N
- `#search=keyword` — highlight search results
- `#page=N&search=keyword` — combined jump + search
- `#zoom=scale` — set zoom level

## Use case

Link notes directly to specific PDF content without copying context into the note itself. For example, a bug report note can link to page 42 of a specification PDF where the regression is documented.

## Testing

1. Place a multi-page PDF in `vault/inbox/`
2. Create a note with `[Test](file.pdf#page=2)` on its own line
3. Verify the embedded iframe opens at page 2
4. Test `![[file.pdf#page=2&search=something]]` embed syntax
5. Cmd/Ctrl+click the link → verify reference pane shows page 2
6. Right-click the link → "Open as Reference" → verify fragment preserved

## Scope

- Only PDF link/embed fragment handling is modified
- No changes to image, audio, video, or other asset types
- No changes to PDF export, PDF creation, or vault management
- `pinnedRefFragment` is runtime-only (not persisted across sessions)

## Changes

| File | What |
|------|------|
| `local-assets.ts` | Add `hrefFragment()` helper; append fragment to PDF iframe src in `buildEmbed()` and anchor href in `enhanceLocalAssetNodes()` |
| `cm-live-preview.ts` | Append fragment to `LocalPdfWidget` iframe src; pass fragment through pin-reference button handlers |
| `store.ts` | Add `pinnedRefFragment` (runtime-only, not persisted); extend `noteRefs` type with optional fragment; accept fragment param in `pinAssetReference`/`pinAssetReferenceForNote` |
| `PinnedReferencePane.tsx` | Read fragment from store; apply to iframe URL and cache keys |
| `Editor.tsx` | Extract and pass fragment on Cmd+click PDF links |
| `Preview.tsx` | Extract and pass fragment in context-menu pin handlers |

### DIFF
--- a/packages/app-core/src/components/Editor.tsx
+++ b/packages/app-core/src/components/Editor.tsx
@@ -25,7 +25,7 @@ import {
   wikilinkHeadingAnchor
 } from '../lib/wikilinks'
 import { openDatabaseFromWikilink, openWikilinkHeading } from '../lib/wikilink-navigation'
-import { classifyLocalAssetHref, resolveAssetVaultRelativePath } from '../lib/local-assets'
+import { classifyLocalAssetHref, hrefFragment, resolveAssetVaultRelativePath } from '../lib/local-assets'
 import { externalLinkUrl, extractLinkAtCursor, resolveInternalNoteHref } from '../lib/internal-links'
 import {
   buildMoveNotePrompt,
@@ -587,7 +587,8 @@ function registerVimCommands(): void {
       if (activePath && vaultRoot) {
         const abs = resolveAssetVaultRelativePath(vaultRoot, activePath, target)
         if (abs) {
-          state.pinAssetReferenceForNote(activePath, abs)
+          const fragment = hrefFragment(target)
+          state.pinAssetReferenceForNote(activePath, abs, fragment || null)
           return
         }
       }
--- a/packages/app-core/src/components/PinnedReferencePane.tsx
+++ b/packages/app-core/src/components/PinnedReferencePane.tsx
@@ -43,7 +43,7 @@ import { slashCommandSource, slashCommandRender } from '../lib/cm-slash-commands
 import { dateShortcutSource } from '../lib/cm-date-shortcuts'
 import { wikilinkSource, wikilinkHeadingSource } from '../lib/cm-wikilinks'
 import { completionNavKeymap } from '../lib/cm-completion-nav'
-import { classifyLocalAssetHref, type LocalAssetKind } from '../lib/local-assets'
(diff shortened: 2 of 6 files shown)


adibhanna:
Thanks for this, @wholetthekfcgo! Reviewed and merged into the `v2.13.2` release branch (we cut releases from a release branch rather than `main`), so it ships in 2.13.2.

Cherry-picked as 5315c91 with your authorship preserved. I verified in the running app that the fragment now reaches the target in every path: markdown link hrefs, inline PDF embeds, the standalone card's inline view, the pinned reference pane, and Cmd/Ctrl+click, with plain fragment-less links unaffected.

One small follow-up I added on top (f22d481): in `PinnedReferencePane`, the fragment was resolved as `noteRef?.fragment ?? globalRefFragment` while the path came from the per-note pin, so a per-note pinned PDF without a fragment could inherit an unrelated global pin's `#page=`. I tied the fragment to the same pin the path comes from.

Closing here since it's already merged via the release branch (it won't auto-close because it targeted `main`). Thanks again for the clean, well-scoped PR and the thorough description.

PR #392, opened 2026-07-14
fix: Excalidraw tab state and Vim tab navigation

junereycasuga:
This fixes a few issues when switching to/from Excalidraw tabs:

- Excalidraw viewport state was not preserved when returning to a drawing.
- Note scroll position could reset after switching away to Excalidraw.
- `gt`/`gT` did not work reliably while focused in Excalidraw.
- After navigating from Excalidraw to a note, focus could remain outside the editor.

The fix preserves Excalidraw viewport state, snapshots note scroll before the editor is torn down, handles Vim tab navigation from Excalidraw, and restores editor focus when landing on a normal note.

---

**Before:**

https://github.com/user-attachments/assets/7dcf86ac-9779-42b8-92d9-7289dbc2fad7

**After:**

https://github.com/user-attachments/assets/94cd12a8-d782-4217-aa4c-0611db63e720

### DIFF
--- a/packages/app-core/src/components/EditorPane.tsx
+++ b/packages/app-core/src/components/EditorPane.tsx
@@ -886,6 +886,23 @@ export function EditorPane({ pane }: { pane: PaneLeaf }): JSX.Element {
     return true
   }, [paneId, setActivePane, setFocusedPanel])
 
+  const rememberCurrentTabScroll = useCallback((path = viewPathRef.current): void => {
+    if (!path) return
+    const prev = recallTabScroll(path)
+    const view = viewRef.current
+    const previewEl = previewScrollRef.current
+    const next: TabScrollPosition = {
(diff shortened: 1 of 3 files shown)


adibhanna:
Thanks for this, @junereycasuga, really clean fix!

I've cherry-picked all three commits onto the `v2.13.2` release branch (your authorship is preserved), so it ships in 2.13.2 rather than landing on `main` first.

Verified end to end: `gt`/`gT` now cycle tabs even from a focused Excalidraw drawing, focus returns to the editor after the jump, and switching to and from a drawing keeps its pan/zoom and the note's scroll position. Build and tests pass on all platforms.

Since it was cherry-picked (different commit SHAs), GitHub won't auto-close this, so I'm closing it manually. Thanks again for contributing!

Issue #272, opened 2026-06-25
Linux (Hyprland/Wayland): default Ctrl+W and Ctrl+Shift+Space shortcuts do not work


P4los:
## Description

On Linux with Vim mode enabled, the default shortcuts **Ctrl+W** (close active tab) and **Ctrl+Shift+Space** (system-wide quick capture) do not work as documented. Changing both bindings to alternative chords works immediately.

### Ctrl+W (close tab)

1. Open ZenNotes on Linux with **Vim mode on** (default).
2. Open any note tab.
3. Press **Ctrl+W**.

**Expected:** Close the active tab (Help → Global shortcuts lists `Mod+W` → Close active tab).

**Actual:** Nothing happens. The tab stays open.

**Workaround:** Rebind "Close active tab" in Settings → Keymaps (e.g. `Mod+Shift+W`), or use `:q`.

### Ctrl+Shift+Space (quick capture)

1. Leave the default quick-capture hotkey (`CommandOrControl+Shift+Space`).
2. Press **Ctrl+Shift+Space** in-app and while another app is focused.

**Expected:** Toggle the floating quick-capture window (system-wide).

**Actual:** Nothing happens with the default binding.

**Workaround:** Change the hotkey (e.g. `Ctrl+Shift+F`) — then it works in-app and globally.

## Environment

| | |
|---|---|
| **OS** | Arch Linux |
| **Compositor** | Hyprland / Wayland |
| **Package** | `zennotes-bin 2.7.0-1` |
| **Vim mode** | On (default) |

## Additional context

- Related: #192, #217, #219
- Hyprland does not bind `Ctrl+Shift+Space` on this machine (only `Super+Shift+Space` for float toggle).
- `xdg-desktop-portal` is running.

## Suggestions

1. Linux + Vim: different default for close-tab on Linux, or a clearer first-run hint that `Ctrl+W` is the Vim pane prefix.
2. Quick capture: change the Linux default and/or warn in Settings when the default chord is unreliable on Wayland.
3. Document `zennotes://quick-capture` for Hyprland compositor bindings.

rtim75:
I beleive same goes for half-page-down and half-page-up: Ctrl+d and Ctrl+u. If I change keymaps to alt+d and alt+u they start working

scottames:
> I beleive same goes for half-page-down and half-page-up: Ctrl+d and Ctrl+u. If I change keymaps to alt+d and alt+u they start working

[...]

Issue #217, opened 2026-06-20
Quick capture (system-wide) not working on arch linux nor ubuntu.


Krasy8:
The system quick capture feature, which by default has key bindings (shift+ctrl+F), doesn't work, doesn't do anything. Tried to change the key binding but with no luck.

I have cloned the repo and with AI-assistance found a workaround, which now gives me a system-wide key binding which opens a floating note even without the zen-notes app running (what I acutally really wonted!).

The below is the summary of what changes have been made to address it:

ZenNotes system-wide quick capture hotkey does not fire on Omarchy/Hyprland Wayland.

Environment:
- OS: Omarchy / Arch-based
- Compositor: Hyprland / Wayland
- Portal services running:
  - xdg-desktop-portal
  - xdg-desktop-portal-hyprland
- `org.freedesktop.portal.GlobalShortcuts` exists on DBus
- ZenNotes version tested: 2.4.0 local build from source

Observed behavior:
- In-app quick capture binding works: leader+q opens the floating quick capture window.
- System-wide quick capture setting is accepted in Settings.
- Electron `globalShortcut.register(...)` reports success.
- But pressing the configured shortcut never invokes the callback, inside or outside ZenNotes.

Relevant log after adding diagnostics:

`[zen:quick-capture] registered hotkey: Shift+CommandOrControl+F`

Expected follow-up log never appears:

`[zen:quick-capture] hotkey pressed: Shift+CommandOrControl+F`

Electron/Wayland fix attempted:
- Added before app.whenReady():
if (process.platform === 'linux') {
  app.commandLine.appendSwitch('enable-features', 'GlobalShortcutsPortal')
}

Result:
- Portal registration path appears active.
- Shortcut still does not fire.
- xdg-desktop-portal-hyprland logs global shortcut activity, but Electron callback is never called.
- 
Working workaround implemented:
- Added a new deep link:
`zennotes://quick-capture`
- Wired it to call the same toggleQuickCaptureWindow() path as the global shortcut.
- Set x-scheme-handler/zennotes to ZenNotes.
- Added Hyprland binding:
`bindd = CTRL SHIFT, F, ZenNotes quick capture, exec, xdg-open zennotes://quick-capture`

This works reliably and opens the floating quick capture window system-wide.

Suggestion:
- Keep trying Electron globalShortcut with GlobalShortcutsPortal for platforms where it works.
[...]

PR #395, opened 2026-07-14
fix(editor): restore Vim half-page shortcuts

scottames:
## Summary

Restore configured Vim half-page shortcuts in the main editor by giving them precedence over CodeMirror’s search and history keymaps.

The bindings still dispatch through `Vim.handleKey`, preserving CodeMirror-Vim’s normal command lifecycle and custom keymap overrides. Insert, visual, and non-Vim modes retain their existing behavior.

## Why

On Linux, CodeMirror interpreted `Mod-d` and `Mod-u` as Ctrl+D and Ctrl+U, consuming those events before CodeMirror-Vim received them. As a result, half-page scrolling worked in Preview but not in Edit or Split.

This reproduced with clean configurations in both source and packaged builds, confirming a source defect independent of a separate conflicting local configuration.

## Scope

This PR intentionally addresses only the Ctrl+D/Ctrl+U behavior reported in a comment on #272.

The other symptoms in that issue use different code paths:

- **Ctrl+W:** In Vim mode, ZenNotes intentionally reserves this as the pane-command prefix while a tab is open; tabs close with `:q`. Changing that behavior or its documentation is a separate product decision.
- **Quick capture:** Ctrl+Shift+Space is registered through Electron’s system-wide shortcut API and is subject to Wayland/compositor behavior. That path is independent of CodeMirror and has a separate `zennotes://quick-capture` fallback from #217/#219.

Keeping these concerns separate limits this change to the reproduced editor keymap-precedence defect and allows each remaining behavior to be evaluated independently.

This addresses the Ctrl+D/Ctrl+U portion of #272; it does not close the entire issue.

## Testing

- Added regression coverage for keymap precedence, configured overrides, and mode deferral
- App-core tests: 738 passed, 1 skipped
- Desktop typecheck and build passed
- Verified Edit, Split, and Preview behavior with an isolated runtime profile

This addresses the Ctrl+D/Ctrl+U portion of #272.

### DIFF
--- a/packages/app-core/src/components/Editor.tsx
+++ b/packages/app-core/src/components/Editor.tsx
@@ -48,6 +48,7 @@ import { navigateActiveBuffer } from '../lib/buffer-navigation'
 import { applyVimInsertEscape } from '../lib/vim-insert-escape'
 import { listContinuationPrefix } from '../lib/list-continuation'
 import { focusEditorNormalMode } from '../lib/editor-focus'
+import { toVimSequence } from '../lib/vim-key-sequence'
 
 let vimCommandsRegistered = false
 let syncedVimBindings: Partial<Record<KeymapId, string[]>> = {}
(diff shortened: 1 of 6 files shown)


adibhanna:
Thanks, @scottames, this is a really well-put-together fix, and I appreciate the clear scoping and the tests.

I've cherry-picked it onto the `v2.13.2` release branch (your authorship is preserved), so it ships in 2.13.2 rather than landing on `main` first.

Verified locally: typecheck, the full app-core suite (including your new keymap-precedence tests), and the desktop build all pass, and `Ctrl+D`/`Ctrl+U` half-page scroll works in the editor here. I don't have a Linux box to reproduce the original `Mod` chord behavior directly, but the precedence tests cover the mechanism and your write-up made the root cause clear.

Since it was cherry-picked (different commit SHAs), GitHub won't auto-close this, so I'm closing it manually. This addresses the `Ctrl+D`/`Ctrl+U` part of #272; I'll leave #272 open for the remaining Ctrl+W and quick-capture items you called out. Thanks again for contributing!

PR #406, opened 2026-07-16
Add custom TextMate code language support

rossirpaulo:
Adds desktop support for importing self-contained TextMate JSON grammars and using their fence tags in both the editor and rendered Markdown.

Users can add, preview, enable, edit, reveal, and remove custom languages from Settings → Editor → Languages. Installed grammars are stored in the machine-wide ZenNotes config directory and update open editor and preview surfaces without changing note files.

The runtime keeps built-in language and diagram tags reserved, rejects grammars with unsupported external scope dependencies, bounds grammar and code-block sizes, and loads TextMate/Oniguruma only when needed. The Oniguruma WASM is embedded and decoded locally instead of fetched, so it works with Electron's CSP and packaged file URLs.

Validation completed:

- 58 shared-domain tests
- 804 app-core tests
- custom-language desktop persistence tests
- TypeScript checks for shared-domain, bridge-contract, app-core, web, and desktop
- web and desktop production builds


rossirpaulo:
I'm adding this feature since I'm trying to move off Obsidian and I build a programming language. It's hard to move off if it means my snippets don't have custom highlighting. Hope this helps and gets merged. Lmk if you need help reviewing.

rossirpaulo:
@adibhanna, any steps that I need to take to get this PR some momentum? lmk if there's anything not documented in `CONTRIBUTING.md`.

adibhanna:
Merged and shipping in 2.20.0. Thank you, this was a genuinely well-built PR and a pleasure to review.

It landed on the `v2.20.0` release branch rather than through this PR, because the branch had moved a long way since you opened it and the base here still pointed at `main`:

- `8f7f680` the merge, credited to you
- `825ee6e` two small follow-ups from review
- `e7a9879` a bundling fix your PR led me to

## What I had to resolve

Four conflicts. Three were "both sides added something in the same place". The fourth was real: your `rehypeCustomCodeLanguages` was wired into the old flat `processor` in `markdown.ts`, which has since been replaced by a `createProcessor()` factory (a Typst math renderer landed in the meantime). It now sits in that chain, still after `rehype-highlight` so a user grammar wins over highlight.js guessing at a tag it does not really know.

## Two follow-ups

**One perf change.** The CodeMirror extension and the rehype plugin are both installed unconditionally, and both walked a whole tree before discovering they had nothing to do: `buildDecorations` ran `syntaxTree().iterate()` over the document on every doc change and every viewport change, and `rehypeCustomCodeLanguages` visited every element of every rendered note. With no grammar installed, which is every existing vault, that work can never produce a decoration. Both now bail on a new `CustomCodeLanguageRegistry.isEmpty` getter. I also swapped one `text-[11px]` for the `text-2xs` token, which is the same 11px but uses the scale.

I did measure the case the guard does not cover, since I was worried about it: a document with 31 custom fences including one 400 lines long, typing inside it, grammar enabled vs disabled. The p50 delta was -0.1ms across three runs. CodeMirror parses lazily to about the viewport, so the walk never reaches the whole document, and your token cache absorbs the rest. Your design was fine; I was wrong to worry.

I deliberately did not reformat your files to match the repo's quote and semicolon style. There is no prettier config or lint script here, the components match `SettingsModal`'s existing style, and rewriting 1,000 lines of your code would buy nothing but conflict risk.

## The thing your PR uncovered

Chasing where `vendor-markdown` was loading from, I found that the app had been fetching and evaluating 2.5MB of mermaid on every single launch, for reasons entirely unrelated to your work: a `manualChunks` rule was hoisting a dynamic-only import into the entry's static graph. Removing it also made mermaid split properly, so a flowchart now pulls three small chunks instead of one bundle with every diagram type. That had been true for many releases and nobody had noticed. So thank you twice.

## Verification

Beyond typecheck and the suite (shared-domain 1162, app-core 1218, desktop 430), I drove the built app over CDP with a toy grammar installed into an isolated config dir: the grammar is discovered and parsed, the editor tokenizes its fence, the reading view tokenizes the same block, Settings lists it under the new Languages sub-tab, and no Oniguruma or TextMate errors reach the console. Separately, with no grammar installed, the 633KB engine chunk and the 81KB `vendor-textmate` chunk are never requested at all.

Closing this in favour of the merge commit. The feature is credited to you in the release notes.

PR #422, opened 2026-07-18
Fix(notation): fixed legacy notation for Nix

pimkoter:
There were still some old "Xorg.NAME was renamed to X" warnings in ./packaging/nix/package-desktop.nix

### DIFF
--- a/packaging/nix/package-desktop.nix
+++ b/packaging/nix/package-desktop.nix
@@ -7,7 +7,6 @@
   wrapGAppsHook3,
   copyDesktopItems,
   makeDesktopItem,
-
   # Electron / Chromium runtime libraries
   alsa-lib,
   at-spi2-atk,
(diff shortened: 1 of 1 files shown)


adibhanna (changes requested):
Thanks for tackling this, @pimkoter — clearing the `xorg.*` deprecation warnings is a welcome cleanup. A few things to sort out before it can go in:

**1. Please drop / split the unrelated changes.** Beyond the notation fix, this PR also does two things I'd like handled separately:

- `release-data.json` rewrites `"hash"` — the `fetchFromGitHub` **source** hash used by `package-server.nix` — for the already-released **v2.14.0** (`sha256-PdTD3o…` → `sha256-C+doF2…`). That's unrelated to the desktop Xorg fix. If you hit a real source-hash mismatch it's worth its own PR with the `nix-prefetch`/`nix build` output so it can be verified; please revert it here.
- `package-desktop.nix` drops the nested `WAYLAND_DISPLAY:+` guard on the ozone flag: `NIXOS_OZONE_WL:+WAYLAND_DISPLAY:+--ozone-platform-hint=auto` becomes `NIXOS_OZONE_WL:+…`, so the hint is now added whenever `NIXOS_OZONE_WL` is set even with no Wayland display. If that's intentional, please call it out and why; otherwise please keep the original guard.

**2. Please keep the diff minimal.** Most of the 142/133 is a whole-file reformat (indentation, list spacing). Trimming it back to just the ~12 `buildInputs` renames makes it easy to review and less likely to collide with the per-release edits to this file.

**3. Verification.** We can't build the Nix packages in CI or locally on our side (no Linux/Nix here), and the desktop derivation is `x86_64-linux`-only, so we can't confirm the flattened names (`libx11`, `libxscrnsaver`, `libxtst`, …) all resolve on the nixpkgs revision your users pin. Could you paste the output of a successful `nix build .#zennotes-desktop` plus the nixpkgs rev? That's the main thing standing between this and a merge.

Really appreciate the contribution — scoped down to just the notation rename plus a build log, this should be quick to land.

PR #432, opened 2026-07-20
fix(nix) fixed legacy xorg syntax

pimkoter:
Screenshot from 2026-07-20 22-14-00
Built completed on nixpkgs rev 753cc8a3a87467296ddd1fa93f0cc3e81120ee46

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -5,39 +5,35 @@
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
   };
 
-  outputs = { nixpkgs, ... }:
-    let
-      systems = nixpkgs.lib.platforms.linux ++ nixpkgs.lib.platforms.darwin;
+  outputs = {nixpkgs, ...}: let
+    systems = nixpkgs.lib.platforms.linux ++ nixpkgs.lib.platforms.darwin;
 
-      forAllSystems = nixpkgs.lib.genAttrs systems;
-    in
-    {
-      packages = forAllSystems (system:
-        let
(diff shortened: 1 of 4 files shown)


adibhanna:
Thanks a lot for digging into the Nix packaging, @pimkoter — and especially for the `services.zennotes` NixOS module, that's a genuinely useful addition. 🙏

I'm going to close this PR as-is, but I'd love to take part of it. Here's the reasoning, channel by channel:

- **`xorg.libX11` → `libX11` (the titled "fix"):** we just cut 2.15.0, and the `nix-update` workflow ran `nix build .#zennotes-desktop .#zennotes-server` on a Nix runner against current `main` — with the existing `xorg.*` syntax — and it built cleanly. So the desktop/server packages already build fine as-is; this change isn't fixing a break on our side.
- **`release-data.json` hash:** `main` is now at 2.15.0 with freshly recomputed, build-verified hashes, so the 2.14.0 line this PR edits is no longer used. (It also differed from the value our recompute step produced.)
- **Formatting:** the alejandra reformat of `flake.nix` / `package-desktop.nix` is a big diff that we'd rather not fold in alongside functional changes — it makes future diffs noisy against the current style.

What I'd genuinely like to merge is the **`server-module.nix`** on its own. Could you open a small, focused PR with just that file (plus the `nixosModules.server` wiring in `flake.nix`)? I'll run it through CI and get it in for the next release, with credit to you.

Really appreciate the contribution — sorry to send you a second lap on the module, but a focused PR will sail through.

PR #426, opened 2026-07-19
editor: fix frontmatter heading styling

junereycasuga:
> **DISCLAIMER**: I'm not sure if this is the intended UI or not. But creating a PR anyway :)

Fix frontmatter styling in the editor.

CodeMirror can tokenize the line before a closing `---` frontmatter fence as a setext heading when there isn't a blank line before the fence. This caused valid frontmatter to render with large heading typography in edit mode.

This makes frontmatter card styling override heading tokens inside the frontmatter block, so both forms render consistently.

Also adds a preview regression test that verifies leading YAML/TOML frontmatter is hidden from rendered markdown.

**Before:**
image

**After:**
image

PR #443, opened 2026-07-22
editor: toggle inline formats off at closing marker

junereycasuga:
**Summary**

 Fixes inline formatting shortcuts so pressing

- Mod+B (Bold)
- Mod+I (Italic)
- Mod+E (Code)
- Mod+Shift+S (Strikethrough)
- Mod+Shift+H (Highlight)

a second time after typing formatted text exits the formatting span instead of inserting another nested markdown marker pair.

 **Changes**

 - Detects empty-cursor formatting toggles immediately before a closing marker.
 - Moves the cursor after the closing marker to “turn off” the active format.
 - Removes an untouched empty marker pair when toggled again immediately.
 - Adds regression coverage for active-format exit behavior and empty-pair cleanup.

**Before**

https://github.com/user-attachments/assets/7cb98bcf-a5c5-4a07-922e-983f78e85057

**After**

https://github.com/user-attachments/assets/a781a0e6-adb8-43a8-aeec-c8c8383cef1a

### DIFF
--- a/packages/app-core/src/lib/cm-format.test.ts
+++ b/packages/app-core/src/lib/cm-format.test.ts
@@ -51,6 +51,29 @@ describe('toggleWrap', () => {
     expect(view.state.selection.main.head).toBe(3) // between == and ==
   })
 
+  it('moves past the closing marker when toggling off an active empty-selection wrap', () => {
+    const view = mount('**bold** text', 6, 6) // cursor before closing **
+    toggleWrap(view, '**')
+    expect(view.state.doc.toString()).toBe('**bold** text')
+    expect(view.state.selection.main.empty).toBe(true)
+    expect(view.state.selection.main.head).toBe(8)
+  })
+
+  it('removes an untouched empty pair when pressing the shortcut again', () => {
+    const view = mount('****', 2, 2) // between the opening and closing **
+    toggleWrap(view, '**')
+    expect(view.state.doc.toString()).toBe('')
+    expect(view.state.selection.main.empty).toBe(true)
+    expect(view.state.selection.main.head).toBe(0)
+  })
+
+  it('does not treat a cursor before a later opening marker as active formatting', () => {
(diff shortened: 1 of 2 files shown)


adibhanna:
Merged into `v2.19.0` — thanks @junereycasuga, this ships in 2.19.0.

Good fix, and the right instinct about where it belonged: pressing the shortcut a second time should get you *out* of the span, not stack another pair inside it. I checked it by driving the real editor rather than only the unit tests — `Cmd+B` `bold` `Cmd+B` ` plain` now writes `**bold** plain`, and the code/strike/highlight markers behave the same way.

One follow-up commit on top ([ed3fe0b](https://github.com/ZenNotes/zennotes/commit/ed3fe0b)), for an edge the tests didn't reach. The empty-pair removal matched on the marker's characters without asking which marker owned them, so in a fresh `**|**` from `Cmd+B`, pressing `Cmd+I` saw a `*` on each side of the cursor, read it as an empty italic pair, and deleted it — you asked for italic inside bold and your bold turned into a broken italic pair. Removal now stands down when a longer marker forms a pair at the same spot, the same longest-marker-wins rule the Backspace handler has used since #468 (which is why `WRAP_MARKERS` is ordered longest-first). `Cmd+B` then `Cmd+I` now nests to `***|***`, while `Cmd+B` inside an empty bold-italic still strips the bold and leaves the italic — so toggling a format off where it genuinely is on still works.

I also pulled the empty-pair check into one `isEmptyPairAt` shared with the Backspace handler, since the two have to agree on what an empty pair is, and gave `isInsideUnclosedMarker` a comment about being a count rather than a parse.

Two things I left alone, in case you want them: a cursor in the *middle* of a formatted span (`**bo|ld**`) still inserts a nested pair rather than unwrapping, and a stray `$` earlier on the line (`costs $5 and $x|$`) throws off the parity count for math. Both predate this PR and neither is a regression.

Third one of yours to land, after #426 and #281. Appreciated.

PR #444, opened 2026-07-22
ui: include frontmatter tags in tags sidebar section

junereycasuga:
**Summary**

This PR makes frontmatter tags a first-class source for note tags, alongside inline #tags.

Previously, the sidebar/tag metadata pipeline only picked up inline hashtag syntax. Notes that already had tags in notes frontmatter, such as:

 ```markdown
---
tags: [project, daily, area/work]
---
 ```

 would not contribute those tags to the sidebar unless the same tags also appeared in the body.

 **Changes**
 - Include tags from frontmatter when extracting note tags.
 - Support common frontmatter tag formats:
     - inline arrays: tags: [daily, work]
     - block lists:
```markdown
tags:
  - daily
  - work
```
     - quoted tags and tags with a leading #
     - nested tags like project/compiler
 - Keep inline #tag extraction behavior unchanged.
 - Ignore non-tags frontmatter fields containing #, e.g. title: #ignored.
 - Bump the note metadata cache version so existing notes are reparsed and old cached metadata does not hide pre-existing frontmatter tags.
 - Keep desktop, renderer, MCP, and server tag extraction paths consistent.
 - Add regression coverage for frontmatter tag extraction.

 **Validation**

 - `npm run test:run --workspace @zennotes/app-core -- tags.test.ts`
 - `npm run test:run --workspace @zennotes/desktop -- vault.test.ts`
 - `go test ./internal/vault`
 - app-core and desktop node typechecks passed.

### DIFF
--- a/apps/desktop/src/main/vault.test.ts
+++ b/apps/desktop/src/main/vault.test.ts
@@ -624,6 +624,32 @@ describe('listNotes metadata parsing', () => {
     expect(note?.tags).toEqual(['realtag'])
   })
 
+  it('indexes frontmatter tags as first-class note tags', async () => {
+    const root = await makeTempDir('zennotes-meta-frontmatter-tags-')
+    await ensureVaultLayout(root)
+    const rel = 'inbox/frontmatter.md'
+    await writeFile(
+      path.join(root, rel),
+      '---\ntags: [frontmatter, "#quoted", project/nested]\ntitle: #ignored\n---\n\n#inline\n',
+      'utf8'
+    )
+
+    const notes = await listNotes(root)
+    const note = notes.find((n) => n.path === rel)
+    expect(note?.tags).toEqual(['frontmatter', 'quoted', 'project/nested', 'inline'])
+  })
+
+  it('indexes block-list frontmatter tags', async () => {
+    const root = await makeTempDir('zennotes-meta-frontmatter-tag-list-')
+    await ensureVaultLayout(root)
+    const rel = 'inbox/frontmatter-list.md'
+    await writeFile(path.join(root, rel), '---\ntags:\n  - daily\n  - "#log"\n---\n\nBody\n', 'utf8')
+
+    const notes = await listNotes(root)
+    const note = notes.find((n) => n.path === rel)
+    expect(note?.tags).toEqual(['daily', 'log'])
+  })
+
   it('detects only local asset references as attachments', async () => {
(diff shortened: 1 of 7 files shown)


adibhanna:
Merged into `v2.19.0` — thanks again @junereycasuga, this ships in 2.19.0 alongside #443.

This one was a genuine gap: a vault written in Obsidian, where tags live in frontmatter, showed an empty tag pane in ZenNotes unless the same tags were repeated in the body. I confirmed it in the packaged app before and after — the seeded vault went from listing `inline`, `notatag` to listing `alpha`, `beta/gamma`, `daily`, `work`, `log`, `inline`. Note that second item in the "before" list: `title: #notatag` in frontmatter used to be scanned as an inline tag, because the block was never excluded from the body scan. Your change fixes that too, which is worth calling out since it's the kind of thing that quietly pollutes a tag list.

Bumping the note-meta cache version so existing vaults reparse was exactly right, and covering all four extraction paths — renderer, main process, MCP, Go server — is more than most PRs here do.

One follow-up commit on top ([6cb27b5](https://github.com/ZenNotes/zennotes/commit/6cb27b5)):

- **One parser instead of four.** The renderer and main-process copies each got a fresh hand-rolled `parseSimpleFrontmatter`, while shared-domain already had the same "just enough YAML" parser behind `parseTaskFrontmatter` and MCP/Go reused theirs. It now lives in `@shared/frontmatter` as `parseFrontmatterFields` + `frontmatterTags`, and the renderer, main process and task parser all call it. MCP and Go stay hand-synced — neither can import the package — but their comments point at the one definition now. This codebase already keeps five copies of the tag scanner aligned by hand; adding two more parsers for the same file format was the thing to avoid.
- **`tags: daily, work` is two tags.** A bare scalar (no brackets, no block list) parsed as the single tag `daily, work`, comma and space included, which the tag pane can never match and which every other editor reading this field treats as two. A bare scalar now splits on commas and whitespace — lossless, since a tag can hold neither. Same rule in all four copies, with tests in app-core and Go.

Nothing wrong with your parsing beyond that scalar case; the block-list, inline-array, quoted and leading-`#` handling all checked out, including CRLF and "another key's block list doesn't leak into tags", which I probed specifically.

Heads up on one place this doesn't reach yet: the iOS app keeps its own copy of the vault semantics in a separate repo, so frontmatter tags won't show there until I port the same change. That's on me, not on this PR.

Issue #534, opened 2026-08-02
Workflows canvas: multi-select nodes (Shift/Ctrl-click, Ctrl+A)


adibhanna:
Split out of #532, which reported it alongside a Cyrillic input bug. The Cyrillic half is fixed in 43d9923; this is the other half, and it is a feature rather than a regression.

**What is asked for**

- `Ctrl/Cmd + A` on the canvas selects every node.
- `Shift` or `Ctrl` clicking nodes builds up a selection instead of replacing it.
- A selection can then be moved (or otherwise acted on) as a group.

**Where the canvas is today**

Selection is deliberately single: `selectedNodeId` is one id, the options panel edits one step, and `onPaneClick` clears it. React Flow's own multi-selection is not enabled. So this is not a broken feature, it is an absent one.

**What needs deciding before building it**

The interesting question is not how to select several nodes, it is what a group is FOR. React Flow will happily paint a marquee, but a selection that can do nothing is worse than no selection.

- **Move** is the natural first action, and the cheapest: a drag already stores a per-node position override, so moving several is the same write repeated. Probably where this should start.
- **Delete** is the one that needs care. Deletions rewrite the workflow text by statement and step index, so removing several at once has to renumber correctly, and a partial failure would leave the file half-edited. The single-node delete path is careful about this for one node.
- The options panel stays single-node either way; there is no sensible "edit the arguments of five different step types at once".

Worth noting the canvas is keyboard-navigable already (Tab walks nodes, Enter opens options, Backspace deletes the selected one), so whatever selection model lands should be reachable from the keyboard too, not mouse-only.

Reported by @ShowhyT in #532.

adibhanna:
Built in 6de3288, shipping in 2.21.0. Thanks @ShowhyT for asking for it in #532.

**What works now**

- `Shift` + click, or `⌘` + click on macOS / `Ctrl` + click elsewhere, adds a step to the selection.
- `Shift` + drag across the canvas draws a selection box over several.
- `Mod+A` on the canvas selects every step. Inside a text box it still selects the text.
- Dragging any member of a group moves the whole group, and every new position is written to the file's layout line together.
- `Esc` clears the whole selection.

**How it stays safe**

Selection is still a single id everywhere that edits: the options panel, delete, and the context menu all ask for "the one node", and a group answers that there is no one node. So nothing that rewrites a step can fire against five of them, and the options panel steps aside while a group is selected rather than editing an arbitrary member.

[...]

Issue #533, opened 2026-08-02
Custom folder path not working correctly


hotondo:
Linked to issue #115 
First of all, thank you for implementing that. 
Unfortunately it does not work as expected, at least on archlinux systems. The gui does not accept custom folders for all folders and does not accept folders that are not on the top level. so a path like `docs/zennotes/quick` for exampe can be put in the `vault.json` config file but only manually, not via the gui. 
If more than one of the 4 is entered via the gui, the already edited path is removed again. If a path such as the example path above is entered it is not accepted.
More serious is that even if it is entered manually in the `vault.json` file, the configuration is not respected. If i reopen zennotes it still creates the defaul quick folder on root level for example and the new quick notes are put in there instead of the path defined in the config file.

adibhanna:
Thank you for the detailed report, and sorry it cost you time. You were right that something is wrong, though not quite the thing it looked like. Partly fixed in 6a7ab3f for 2.21.0, and partly a limit I want to be straight with you about.

**The short version:** a system folder is one directory name at the top level of your vault, never a nested path. `docs/zennotes/quick` was never going to work. What was broken is that nothing said so: the box went blank, the file kept your line, and the app carried on with the default, which is indistinguishable from a broken feature.

**Taking your three points in turn.**

**1. Nested paths are refused.** By design, and the design has a reason. Every part of the app that asks "which system folder is this note in?" answers by looking at the first segment of its path, and that rule exists in three places that must agree: the shared TypeScript, a copy inside the MCP server, and a Go mirror in the self-hosted server. Allowing `docs/zennotes/quick` changes what "the folder a note belongs to" means in all three, and in every vault already on disk. It is a real feature request, not a validator that needs loosening. If you want it, open an issue and say what your layout looks like, because the shape of the ask matters for whether it can be done safely.

**2. A hand-edited vault.json is ignored.** Same cause, and this is the one I would also have called serious. Your line stayed in the file, so it looked respected; the loader dropped it and fell back to `quick/`. Now, when a stored path cannot be used, the app logs which line it ignored and why, once per change to the file. Your original line would have produced:

```
[zen] vault.json: systemFolderPaths.quick = "docs/zennotes/quick" was ignored.
Use a single folder name at the top level of your vault, not a path.
```

[...]

Issue #532, opened 2026-08-02
[BUG] Workflows Canvas mode: Cyrillic input issues and missing selection shortcuts (Ctrl+A, Shift/Ctrl)


ShowhyT:
### **Description**

Several issues regarding text input and selection shortcuts occur in Workflows **Canvas** mode:

1. **Cyrillic Input Issue:** When editing nodes or parameters in Canvas mode, entering Cyrillic characters is unstable or fails to register in certain fields (for example, when attempting to input a tag or search string with Cyrillic characters).
2. **Missing Selection Shortcuts:**
* `Ctrl + A` does not select all nodes/notes on the canvas.
* Selecting multiple elements by holding `Shift` or `Ctrl` while clicking on nodes does not work.

---

#### **1. Cyrillic Text Input:**

1. Open **Workflows -> Canvas** mode.
2. Create or open a workflow containing search or filtering logic with Cyrillic characters (e.g., using tag `#кириллица-работает` or searching for a note title).
3. Try to type or edit Cyrillic text directly in the node parameters inside the Canvas interface.
4. **Result:** Cyrillic characters are either partially entered, ignored, or break the filtering logic in Canvas mode.

#### **2. Selection Key Combinations:**

1. While on the Workflows Canvas, press `Ctrl + A`.
* **Result:** Nothing happens (nodes/notes are not selected).

2. Try to select multiple notes/nodes by holding down `Shift` or `Ctrl` and clicking on individual nodes.
* **Result:** Multi-selection fails, resetting any previous selection.

---

### **Expected Behavior**

* Cyrillic character input across all text fields on the Canvas should function as reliably as Latin characters.
* Pressing `Ctrl + A` on the Canvas should select all active workflow elements/nodes.
* Holding `Shift` or `Ctrl` while clicking on nodes should allow selecting a group of elements for bulk movement or actions.

---

### **Sample Workflow to Reproduce**

```yaml
---
name: Cyrillic Support Check
description: Tests Cyrillic text processing in Canvas mode
status: active
trigger: manual
---

folder2 = folder "inbox/Cyrillic Test" | contains #кириллица-работает | append Tag search with #кириллица-работает not work in Canvas Mode
search2 = search Cyrillic Test | where title matches Кириллица заметка | append Search in note body using Cyrillic
tag2 = tag #кириллица-работает | prepend Cyrillic tag editing in Canvas not work
```

adibhanna:
[...]

Issue #530, opened 2026-08-02
Feature Request: Render mermaid code blocks inline in editor mode


junereycasuga:
Currently, mermaid diagrams are only rendered in preview/split mode. In the editor they remain plain `mermaid` fenced code blocks, so authors must switch to preview mode to see the diagram.

**Request:** render mermaid code blocks inline in the editor.

**Behavior:**
- When the cursor is outside the code block, show the rendered mermaid diagram in place of the code block.
- When the cursor moves inside the code block, switch to the raw code view so the user can edit the source.

This matches the pattern used for other rich embeds in the editor and would make authoring diagrams more convenient without leaving edit mode.

Add inline rendering for mermaid blocks in the editor:
- Cursor outside the block → render the diagram.
- Cursor inside the block → show the source code for editing.

This would let users author and verify diagrams without switching to split/preview mode.

naingyeminn:
I’m supportive of this idea.

adibhanna:
Built in 4737d92, shipping in 2.21.0. Thanks for the clear spec, @junereycasuga, and @naingyeminn for the second vote.

It behaves exactly as you described: cursor outside the block draws the diagram, cursor inside brings the source back, and moving away redraws it. Live preview has to be on, since this is part of that bargain rather than a separate mode.

**The part worth reporting back**, because it nearly made the feature useless: a rendered block is a single widget with no cursor positions inside it, so CodeMirror's vertical motion sails straight over it. In my first working build, pressing down-arrow twelve times never once landed in the block, and only a mouse click could open the source. In an app whose identity is the keyboard, that is a trap rather than a feature. The block now publishes its line ranges the way rendered math and embeds already do, and the two motion helpers that knew about those learned about this one, so arrows and Vim `j`/`k` step in properly.

Verified in the built app, editor mode, both with and without Vim:

| | result |
|---|---|
| cursor elsewhere | diagram drawn, labels correct, source hidden |
| arrow-down / `j` toward it | lands inside on the 4th press, fence returns as text |
| click the diagram | cursor lands in the block, source revealed |
| move away again | redraws |
| theme switch | redraws in the new palette |

Three details you did not ask about but will notice:

- **A diagram mid-edit keeps its last good drawing** rather than flashing a parse error at every keystroke. The error only takes over once you leave the block alone and it still does not parse, which is the difference between a diagram you are typing and one that is wrong.
[...]

PR #535, opened 2026-08-03
ZenNotes 2.21.0

adibhanna:
The accumulating branch for 2.21.0. Opened as a draft: the cycle is still open, so this grows as work lands, and `main` fast-forwards here at release time rather than merging.

13 commits, 62 files, one outside contribution.

## New

- **A task can say it is in progress** (`a65b119`, #512). `- [/]` joins open, done, forwarded and cancelled, following the Obsidian Tasks convention. It differs from the other two non-empty states in the way that matters: in progress is still open work, so it stays in Today, on the calendar and on the board, and it rolls forward with the unfinished tasks when a daily note rolls over. Set it by typing it, from the palette, or with `i` in the list, on the board or on the calendar. Round-trips through the canonical parser, the MCP scanner, the `zn` CLI and the Go server. Asked for by @NunonuN.

- **Right-click actions on every task surface** (`a65b119`). The Tasks list, the Kanban board, the Tasks calendar and the calendar side panel now share one menu builder: open, the four state changes, waiting, forward, due presets, priority, delete, each showing the key that does the same thing. Before this the calendar had a menu and the list and board had none.

- **Mermaid diagrams draw while you write** (`4737d92`, #530). A ```mermaid fence renders in the editor, not only in preview; the cursor entering the block brings the source back. Arrow keys and Vim `j`/`k` step into a drawn diagram, so the source is reachable without the mouse. Mermaid only, deliberately. Asked for by @junereycasuga, seconded by @naingyeminn.

- **Multi-select on the Workflows canvas** (`6de3288`, #534). Shift-click, Shift-drag a box, or `Mod+A`, then drag any member to move the group. Bulk delete is deliberately out: it has to renumber the file as it goes. Asked for by @ShowhyT.

## Fixes

- **`zn` runs again on a packaged install** (`186c6df`, #524). Dead since 2.20.0 with `Cannot find module 'smol-toml'`, on every platform and every install method, because the CLI ships beside the app in a folder with no library directory and one dependency was left to be resolved at startup. A development checkout resolves it from the workspace above, which is why three releases shipped a CLI nobody could run. There is now a build-time check that runs the built CLI somewhere with nothing around it. Reported by @oliviertripet, confirmed by @naingyeminn.

- **Wikilinks into a database resolve on a remote vault** (`4e9f8d3`, #527). Same vault, same link, different answer depending on whether the files were read locally or through the server; clicking through would have created a duplicate note. Reported by @naingyeminn.

- **The Workflows tutorial cleans up after itself** (`75869da`, #525). It promises the practice folder and its runs disappear; on some vaults the folder stayed. Reported by @oliviertripet.

- **Cyrillic tags survive being typed on the canvas** (`43d9923`, #532). The per-keystroke tag check knew only the English alphabet, so `#кириллица-работает` collapsed to `--` as you typed. The rule for what a tag may contain now lives in one place. Reported by @ShowhyT.

- **The CLI install screen reads the real PATH** (`fa4944f`, #528). It offered to add a directory that was already there.

- **A refused folder path says why** (`6a7ab3f`, #533). Settings ▸ Vault ▸ Folders silently blanked the box on an invalid path, and a hand-edited `vault.json` line was ignored without a word. Nested paths remain unsupported on purpose. Reported by @hotondo. The issue stays open: one of the three reported symptoms did not reproduce.

- **The dock icon loads on an installed Mac app** (`92feee2`). Cosmetic, and wrong since 1.1.0.

- **Four task-surface bugs found while building `[/]`** (`a65b119`), all older than it: `[-]` and `[>]` rendered as literal brackets in the reading view, in HTML and PDF export, and in shared notes, contradicting what the manual has claimed since 2.19; clicking a checkbox in the reading view wrote to the wrong line when a stateful task sat above it; editing a forwarded task on either calendar saved it back stripped of its `due:`, `!priority`, `#tags` and backlink; and the list and board drew every state as the same empty checkbox. The cursor row also advertised "Space check", which cannot fire because Space is the Vim leader.

## Contributions

- **#529, mermaid labels clipped in preview**, by @junereycasuga (`9ee84c8`, merged as `85b43b1`). Node text was measured against a font that had not loaded. Follow-up in `dd6bc46` moves the font fallback into one place rather than leaving the same rule in two files.

## Verification

`npm run typecheck` clean. 3,141 tests pass across shared-domain, app-core and desktop, plus `go vet` and `go test`. The task work was additionally driven over CDP against the built app with isolated `ZENNOTES_USER_DATA_PATH` and `ZENNOTES_CONFIG_DIR`: 15/15 checks for the state itself (editor marker, reading-view markers, the corrected checkbox indexing, the Tasks row, `i` writing to disk) and 16/16 for the menus across list, board and calendar. A demo clip is recorded under `docs/releases/v2.21.0/media/`.

## Still open, deliberately

- **The second half of #512**: rolling a parent task up automatically when some of its subtasks are done. It turns on whether the parent's line should be rewritten on disk (survives into other tools, but only ZenNotes' own writes would fire it, so a subtask checked in Obsidian or on the phone leaves the file contradicting itself) or the progress simply shown. Asked back on the discussion; not built.
- **#526**, hint mode not activating the Edit button. Investigated, not reproduced, waiting on the reporter. Not claimed anywhere in the release copy.

Website docs for the task work are already pushed (ZenNotes/website `ab2546c`); the in-app manual half is in this branch.


adibhanna:
The Windows job on this branch was red before any of this cycle's work: `TestReadNoteStatusesDistinguishMissingFromBroken` has been failing on `main` since `ddfb63d` (July 31), and only on Windows.

The server classified "you asked me to read a directory as a file" from the errno, testing for `EISDIR`. That is the Unix answer; Windows returns `ERROR_INVALID_FUNCTION` ("Incorrect function" in the CI log), so the check missed and the request fell through to a 500 where macOS and Linux said 400. Same request, two answers, depending on the platform.

Fixed in `a713ee6`. `ReadNote` already stats the path before reading it, so it now returns a `vault.ErrIsDirectory` sentinel from that stat, in the same shape as the `vault.ErrPathEscape` the handler already maps. No errno involved, so there is nothing left to differ by platform. The `EISDIR` branch stays underneath as a fallback for read paths that have not been classified.

The end-to-end test only fails on Windows, which is a poor place to keep a guarantee, so the classification is also pinned in the vault package by a test that means the same thing on every runner.

Issue #570, opened 2026-08-10
[BUG] Cannot resize attached images whatsoever. Does not respect both obsidian's format and zennote's


corpserot:
ZenNotes v2.26.0

I tried this, following suggested excalidraw format

```md
![[assets/cognitive_web.jpg|100x50]]
```

I tried this, following obsidian's format

```md
![cognitive web|100x50](../../assets/cognitive_web.jpg)
```

both resulted in this, showing that zennotes does not respect specified image size constraints when displayed both in editor and preview (not shown, cuz it's the same thing).

<img width="1366" height="738" alt="Image" src="https://github.com/user-attachments/assets/ef9d943c-1468-4d19-81ff-dea8e8f39cd8" />

i cannot select text in help, you can read it [yourself in the source code](https://github.com/ZenNotes/zennotes/blob/9dfed19a548c1649939e5b12cb076e06e14dc7f9/packages/app-core/src/lib/help.ts) about what your LLM wrote in regards to image (and excalidraw) sizes. (btw, that file looks horrible... definitely shaking my faith in LLM discipline used in this project)

adibhanna:
You are right, and the help was writing checks the renderer never cashed: size hints were documented as "same syntax as images" but only Excalidraw embeds ever honored them. Plain images ignored the hint entirely and leaked it into the alt text, in both spellings you tried.

Fixed in 302a5ff on the `v2.27.0` branch. Both of your exact lines now work:

- `![[assets/cognitive_web.jpg|100x50]]` renders at 100x50, and `![[assets/pic.png|300]]` sets width with height following the aspect ratio.
- `![cognitive web|100x50](../../assets/cognitive_web.jpg)` renders at 100x50 and keeps "cognitive web" as the caption. Pipes inside a caption survive; only a trailing segment that parses as a size is consumed.
- Applies in the editor's live preview, the reading view, and the PDF export path (same renderer). A hint wider than the pane is still capped by the layout's max-width.

Worth admitting: the first attempt at this fix set the width/height attributes and looked correct in the HTML while the pixels on screen did not move, because the embed stylesheet stretches images to the pane width and attributes lose to CSS. Measuring the rendered size in the running app caught it; the hint now also lands as inline style. The verification run uses your exact two lines and asserts `getBoundingClientRect` comes back 100x50 in both Edit and Preview.

[...]

Issue #569, opened 2026-08-10
[Feature Request] Don't make :w and :wa close tabs


simonlearnscoding:
As a Vim user I have a strong muscle memory to :wa after any change and it closing active Tabs messes with my workflow as that is not the expected behavior coming from vim please don't let it behave like this or make it a toggle-able setting to choose between :wa and :wqa I'm sure I'm not the only one constantly tripping over this

adibhanna:
You are right, and this was simply a wrong reading of the vim contract on our side. `:wa` had been registered as a plain alias of `:qall` on the theory that autosave leaves nothing to write, so the "write all" half was treated as a no-op and the "close everything" behavior leaked in. Your muscle memory is the correct spec: `:wa` is a save command, full stop.

Fixed in aa0c262 on the `v2.27.0` branch, with vim semantics as the only behavior (no toggle needed):

- `:wa` / `:wall` now write every unsaved note across all panes and close nothing.
- `:wqa` / `:wqall` is newly registered as the "save everything and quit everything" spelling, alongside the existing `:xa` / `:xall`.
- `:qa` / `:qall` / `:quitall` keep closing everything, `:w` stays write-only, `:q` / `:wq` still close just the active note.

The in-app help and website docs listed `:wa` in the close-everything row; both now document it as "write every unsaved note".

How to test locally once it ships (or from the branch): with vim mode on, open a few tabs, make an edit, type `:wa`. Before: every tab closes. After: nothing closes and the edit is on disk (the status quo ante for your workflow). `:wqa` or `:xa` when you actually want the close-everything behavior.

Ships in 2.27.0. Thanks for the clear report.

adibhanna:
Closing as fixed by aa0c262 on the v2.27.0 branch; ships with 2.27.0.

Issue #568, opened 2026-08-09
Vim `f` Motion Triggers Context Menu in Visual Mode


uNyanda:
- **Date:** 2026-08-09
- **Type:** Bug
- **Severity:** Low

#### Environment
- **OS:** Linux
- **WM:** Niri
- **Compositor:** Wayland
- **Software version:** 2.26.0

## Description

When using Vim **visual mode**, the `f{char}` motion conflicts with the context menu shortcut.

For example, pressing `fm` should find and highlight the next `m` character. Instead, the `m` key triggers the context menu.

## Steps to reproduce

1. Open a note.
2. Press `v` to enter **visual mode**.
3. Press `f`.
4. Press `m`.

## Expected Behaviour

The pending `f` motion should have higher priority than the context menu shortcut.

- Pressing `fm` should find and highlight the next `m` character.
- Pressing `m` normally, without a pending `f` motion, should continue to trigger the context menu.

## Actual Behaviour

When `f` is pressed in **visual mode**, pressing `m` triggers the context menu instead of being interpreted as the target character for the pending `f` motion.

## Additional Notes

After `fm` triggers the context menu, pressing `Escape` correctly closes the context menu. However, the pending `f` motion appears to remain active.

For example:

1. Press `v` to enter **visual mode**.
2. Press `m`, which opens the context menu.
3. Press `Escape` to close the context menu.
4. Press `h`.

The `h` is then treated as the target character for the previous `f` motion, causing the highlight to move to the next `h`.

This suggests taht the pending `f` motion is not being cleared when the context menu interaction occurs.

### Suggested Solution

Give a pending Vim `f` motion higher input priority than the context menu shortcut in **visual mode**.

When `f` is awaiting a target character, the next character should be consumed by the Vim motion handler. This would allow `fm` to find and highlight the next `m`, while pressing `m` normally would continue to trigger the context menu.

This preserves the existing context menu shortcut while allowing `f{char}` motions to work correctly.

adibhanna:
Thanks for the unusually precise report: the leak observation (Escape after the menu, then `h` getting swallowed as the f target) was the detail that pointed at the real layer.

What happened: the visual-mode `m` context-menu shortcut lives in a window-level keydown listener registered with capture, so it runs before CodeMirror's vim plugin ever sees the event and stops propagation. Vim had already gone pending on `f`, the `m` was intercepted for the menu, and the still-pending motion then consumed the first key after the menu closed. Both symptoms, one interception.

[...]

Issue #566, opened 2026-08-09
Auto-pair brackets option and VARIABLES lead to four closing brackets


ryanthara:
When editing a template with the activated Auto-pair brackets option, additional two closing brackets are inserted every time, when using VARIABLES which are enclosed by design in double brackets.

Example: `{{cursor}}` is completed to `{{cursor}}}}`

adibhanna:
Thanks for the report, and for naming the exact interaction: it is the combination of the two features, not either one alone.

What happened: with auto-pair brackets on, typing `{{` gives you `{{}}` with the caret in the middle, which is correct on its own. But the template variable completion then replaced only the text from the opening braces to the caret with the full `{{cursor}}`, so the auto-paired `}}` sitting after the caret survived, and every accepted variable came out as `{{cursor}}}}`. (Typing the braces out by hand was never affected: typing `}` over an auto-paired closer steps over it. Only accepting a variable from the popup left the extras behind.)

Fixed in f872b5b on the `v2.27.0` branch: the completion's inserted text carries its own closing braces, so applying it now swallows up to two `}` characters already at the caret. That also fixes a cousin you could hit without auto-pair: editing an existing variable in place (say `{{date}}` down to `{{ti}}`) and accepting `{{time}}` used to leave the old pair's braces behind too.

How to test locally once it ships (or from the branch): with Settings > Editor > Auto-pair brackets on, open Settings > Templates > New template, type `{{cur` and accept the completion. Before: `{{cursor}}}}`. After: `{{cursor}}`. Ships in 2.27.0.

adibhanna:
Closing as fixed by f872b5b on the v2.27.0 branch; ships with 2.27.0.

PR #571, opened 2026-08-10
v2.27.0: jump straight to a tab, plus five fixes

adibhanna:
Release branch for 2.27.0: one feature and five fixes, all issues closed. Per the usual flow, main fast-forwards from this branch at release time rather than merging through the button.

## Feature

- **Jump straight to a tab with {count}gt and Alt+1-9** (#497). {count}gt is absolute, matching vim's own contract (2gt is tab 2, a count past the end lands on the last tab, plain gt still cycles), while {count}gT stays relative. With vim off, nine rebindable shortcuts (tabs.select1 through 9) work from anywhere: Alt+digit cross-platform, Ctrl+digit on macOS because Option+digit types characters on many layouts.

## Fixes

- **Image size hints actually resize the image** (#570). Both `![[assets/pic.jpg|100x50]]` and `caption|100x50` now render at the hinted size in live preview and the reading view, instead of full-width with the hint leaking into the alt text.
- **:wa writes all notes instead of closing every tab** (#569). :wa/:wall were aliased to :qall; they now flush every dirty note and close nothing, matching the vim contract.
- **Pending f/t/r motions own the next key over the m menu** (#568).
- **Template variable completion stops doubling closing braces** (#566).
- **Saved tabs paint before the vault scan finishes** (#564).

Each change was verified over CDP against the built app, and docs are updated in the mirrored surfaces (in-app help, shared keymaps catalog, website rows).


adibhanna (commented):
Multi-agent review of this branch at xhigh effort: 31 candidate findings, 15 survived adversarial verification (14 confirmed, 1 plausible), 2 refuted. 13 are anchored as inline comments; two sit on pre-existing code outside the diff hunks, so they live here:

**correctness / CONFIRMED**: `VimNav.tsx:612`: #497 gave {count}gt absolute-jump semantics, but count-awareness only exists in the editor-focused codemirror-vim path; VimNav's global gt/gT fallback (preview mode, non-editor focus) drops the typed count and cycles one tab instead of jumping to tab N. With 5 tabs open and focus in the rendered preview (the #321 fallback path), typing 2gt moves to the NEXT tab instead of tab 2. The new help.ts entry promises the vim contract without qualifying that it only works while the editor owns focus.

**correctness / CONFIRMED**: `Editor.tsx:773`: the documented ex-line equivalents :tabnext/:tabprevious (and :bn/:bp) still discard any count or argument, so the two spellings of the same vim command now disagree. `:tabnext 3` (or `:3tabnext`), which in vim jumps to tab 3, ignores its params entirely and calls `navigateActiveBuffer(state, 1)`, silently advancing one tab; the same dropped-count inconsistency as the VimNav fallback, but on the ex-command entry point.

The findings cluster in three areas: the #564 workspace-restore rework (phantom/un-prunable tabs), the #497 tab shortcuts (modal bleed-through, override shadowing, AZERTY and Alt-code and Mission Control conflicts, dropped counts), and #570 size-hint parsing (numeric captions consumed, zero dimensions mishandled). Plus one template-completion change that can delete literal braces from note content.

packages/app-core/src/store.ts
```diff
@@ -2436,8 +2440,8 @@ function sanitizeWorkspaceLayout(raw: unknown, existingPaths: Set<string>): Pane
   }
 
   const sanitizePath = (value: unknown): string | null => {
-    if (typeof value !== 'string') return null
-    return existingPaths.has(value) || isWorkspaceVirtualTabPath(value) ? value : null
+    if (typeof value !== 'string' || !value) return null
+    return value
```
adibhanna, line 2444: **correctness / CONFIRMED**: #564: `sanitizeWorkspaceLayout` no longer verifies tab paths, and `refreshNotes`' selectedPath exemption then keeps a never-verified ghost tab alive, so a stale snapshot restores to a phantom active tab with a blank editor instead of being pruned. (Same root cause surfaces at store.ts:4290 and store.ts:4340.)

Vault synced between two machines; the notes open in the saved workspace were renamed or deleted on the other machine. On launch, the active tab's eager read fails and `rewritePathsInTree` prunes it, promoting a sibling stale tab to activeTab (`nextActive = nextTabs[0]`) whose content was never loaded and whose note does not exist. When the index lands, `refreshNotes`' keep() retains that path because it equals `s.selectedPath` (the don't-blank-mid-save exemption), so the user lands on a tab titled after a nonexistent note showing an empty editor that survives subsequent refreshes until manually closed. Pre-2.27 restore ran after `refreshVaultIndexes` and checked every tab against `existingPaths`, so the phantom never appeared.

adibhanna, line 2444: **correctness / CONFIRMED**: With path validation removed here, an all-stale snapshot whose active tab is virtual trips the #384 preserve guard, so every dead note tab is preserved on every refresh for the whole session.

A restored snapshot has a virtual tab (e.g. Tasks or a diagram tab) active plus background note tabs synced from another machine that do not exist locally. The eager read skips background tabs, so they all survive restore. On `refreshNotes` the prune would leave zero note tabs, so `preserveLayoutIfPruneEmptiesNoteTabs` (pane-layout.ts:292) returns the pre-prune layout, and it does so again on every subsequent refresh: the phantom tabs persist all session, and clicking any of them yields a failed note load instead of the tab quietly disappearing as the pre-PR `existingPaths` filter guaranteed.

packages/app-core/src/lib/cm-template-variables.ts
```diff
@@ -22,6 +23,28 @@ export const TEMPLATE_VARIABLES: TemplateVariable[] = [
...
+ * completion option so it can be tested without a mounted editor.
+ */
+export function templateVariableApplySpec(
+  state: EditorState,
+  from: number,
+  to: number,
+  insert: string
+): TransactionSpec {
+  const after = state.doc.sliceString(to, to + 2)
+  const consumed = after.startsWith('}}') ? 2 : after.startsWith('}') ? 1 : 0
```
adibhanna: **correctness / CONFIRMED**: Completion apply no longer replaces exactly the typed range: it unconditionally swallows up to two closing braces after the caret even when they are literal note content.

The old apply dispatched changes for exactly [from, to]; the new `templateVariableApplySpec` always extends the replacement over one or two `}` characters sitting after the caret, without checking they were produced by auto-pair or belong to this variable. In a note documenting escaped template syntax, e.g. the user types `{{da` immediately before an existing literal `}}` that closes a different construct (a Handlebars/Jinja snippet written as prose), accepting the completion deletes those two literal brace characters from the note body, corrupting the surrounding text instead of only inserting the variable.

packages/app-core/src/lib/markdown.ts
```diff
@@ -502,6 +502,32 @@ function remarkTaskMetadata() {
...
+      const image = node as unknown as AnyNode
+      const url = String(image.url ?? '')
+      const isRemote = /^(https?:|data:)/i.test(url)
+      if (!isRemote && classifyLocalAssetHref(url) !== 'image') return
+      const { alt, size } = splitEmbedLabel(typeof image.alt === 'string' ? image.alt : '')
+      if (!size) return
+      image.alt = alt
+      const data = (image.data ??= {}) as { hProperties?: Record<string, unknown> }
+      const hProperties = (data.hProperties ??= {})
+      if (size.width) hProperties.width = size.width
```
adibhanna, line 534: **correctness / CONFIRMED**: #570: falsy-zero width/height checks run after `splitEmbedLabel` has already consumed the label, so a `|0` or `|0xH` hint destroys the alt/caption without applying the corresponding dimension (and `|0x300` applies height only, distorting the image).

A note containing `![[diagram.png|0x300]]` renders in reading view with height="300" but no width attribute (`parseEmbedSizeHint`'s `^(\d+)` accepts 0, `splitEmbedLabel` returns {width: 0, height: 300}, then `if (size.width)` skips the 0), stretching the image, and the label is gone (alt=""). `![[diagram.png|0]]` loses its label entirely with no resize at all. The same falsy-zero pattern repeats in cm-live-preview.ts `LocalImageWidget` (`if (this.width)`) and local-assets.ts (`hintWidth > 0`), so live preview mirrors the distortion.

packages/app-core/src/lib/excalidraw-preview.ts
```diff
@@ -21,6 +23,26 @@ export function parseEmbedSizeHint(hint: string | null | undefined): EmbedSize |
...
+ *  with no hint. Pipes inside the caption survive; only a LAST segment that
+ *  parses as a size is consumed. (#570) */
+export function splitEmbedLabel(label: string | null | undefined): {
+  alt: string
+  size: EmbedSize | null
+} {
+  const raw = (label ?? '').trim()
+  if (!raw) return { alt: '', size: null }
+  const wholeSize = parseEmbedSizeHint(raw)
+  if (wholeSize) return { alt: '', size: wholeSize }
```
adibhanna: **correctness / CONFIRMED**: #570: `splitEmbedLabel` treats a purely numeric markdown alt text as a whole-label size hint, so an intentional numeric caption is consumed: the image is resized and its caption/alt is silently erased.

A note with ``, where '2024' is the caption (a year), now renders the chart with width="2024" (stretched to the pane's max-width cap) and alt=""; the reading-view figcaption falls back to the filename and screen readers lose the alt text. Before 2.27 the caption displayed '2024' and the image kept its natural size. This may be intended Obsidian parity for the wikilink form `![[img|600]]`, but it also rewrites the semantics of pre-existing standard-markdown alt text in users' notes; worth a deliberate verdict.

packages/app-core/src/App.tsx
```diff
@@ -696,6 +702,15 @@ function App(): JSX.Element {
         state.setWordWrap(!state.wordWrap)
         return
       }
+      // Alt+1..9 (⌃1..9 on macOS) — jump straight to tab N (#497). Position
+      // counts across panes in the same order gt cycles through.
+      for (let i = 0; i < TAB_SELECT_KEYMAP_IDS.length; i += 1) {
+        if (matchesShortcut(e, overrides, TAB_SELECT_KEYMAP_IDS[i])) {
```
adibhanna: **correctness / CONFIRMED**: #497: the new tabs.select1-9 shortcuts run in the global window keydown handler with no modal/overlay guard, so they switch the underlying tab while a modal, palette, or the Settings keymap recorder is open.

On macOS with the command palette or Settings open, pressing Ctrl+1 (the default mac binding) preventDefaults and switches the active tab underneath the overlay; when the user closes the overlay they are on a different note than they left, with no visible cause. It also fires while recording a new binding in Settings > Keymaps, changing the tab behind the dialog mid-capture. The repo's stated review rule is that key handlers bail when a modal or menu is open; unlike VimNav (which checks searchOpen/settingsOpen/commandPaletteOpen/etc.), this handler has no such gate.

packages/app-core/src/App.tsx
```diff
@@ -696,6 +702,15 @@ function App(): JSX.Element {
         state.setWordWrap(!state.wordWrap)
         return
       }
+      // Alt+1..9 (⌃1..9 on macOS) — jump straight to tab N (#497). Position
+      // counts across panes in the same order gt cycles through.
+      for (let i = 0; i < TAB_SELECT_KEYMAP_IDS.length; i += 1) {
```
adibhanna: **correctness / CONFIRMED**: The tab-select loop was inserted mid-handler, so the nine new default bindings silently shadow any existing user override on the same keys for shortcuts checked later in the handler (exportNotePdf, zoom, history, panel toggles, etc.).

A user who had rebound a later-checked shortcut to Alt+3 (or Ctrl+3 on macOS) before upgrading, e.g. global.zoomIn or global.exportNotePdf, updates to 2.27.0: the new tabs.select3 default matches first, preventDefaults, and switches tabs, so their long-standing custom shortcut stops working with no conflict warning, because the collision was created by a new default rather than by an edit in the keymap settings UI.

packages/app-core/src/lib/keymaps.ts
```diff
@@ -395,6 +404,23 @@ const KEYMAP_DEFINITIONS: KeymapDefinition[] = [
...
+  // Ctrl+Tab).
+  ...([1, 2, 3, 4, 5, 6, 7, 8, 9] as const).map(
+    (n): KeymapDefinition => ({
+      id: `tabs.select${n}` as KeymapId,
+      kind: "shortcut",
+      scope: "app",
+      group: "global",
+      title: `Go to tab ${n}`,
+      description: `Jump straight to tab ${n}, counted across panes in the same order gt cycles.`,
+      defaultBinding: `Alt+${n}`,
```
adibhanna, line 422: **correctness / CONFIRMED**: The new default Alt+1..9 bindings also match Alt+Numpad digits (`resolveKeyFromEvent` maps Numpad digits to the bare digit at keymaps.ts:1277), hijacking Windows Alt-code character entry.

A Windows user typing an Alt code in the editor (hold Alt, type 0233 on the numpad to get an accented character, a standard Windows input method) now fires tabs.select2 and tabs.select3 mid-entry: App.tsx preventDefaults each digit and switches the active tab twice, so the character entry is broken and the user is yanked to a different note while typing. The Numpad-to-digit resolution predates this PR, but shipping Alt+digit as a default binding is what makes it fire.

adibhanna, line 422: **correctness / CONFIRMED**: Alt+digit default bindings match on `event.key`, so on keyboard layouts where digits are shifted (French/Belgian AZERTY, Czech) Alt+1..9 produce keys like `&` or `é` and the new tab shortcuts never fire.

A Windows or Linux user on a French AZERTY layout presses Alt+1 to jump to tab 1 (as documented in the in-app help): `resolveKeyFromEvent` takes the typed-character fast path and builds "Alt+&", which never equals the stored binding "Alt+1", so nothing happens and the entire #497 non-vim feature is dead by default on those layouts. The codebase already documents this exact trap class (#514, the defaultBindingMac comment about Option+printable chords) but only shielded macOS; the cross-platform Alt+digit defaults introduced here are the first digit-based defaults and hit the same key-vs-layout pitfall unshielded.

packages/app-core/src/lib/keymaps.ts
```diff
@@ -395,6 +404,23 @@ const KEYMAP_DEFINITIONS: KeymapDefinition[] = [
...
+      defaultBindingMac: `Ctrl+${n}`,
```
adibhanna, line 423: **correctness / PLAUSIBLE**: The macOS default Ctrl+1..9 collides with Mission Control's 'Switch to Desktop N' shortcuts, which macOS auto-enables once a user has multiple Spaces, so the OS consumes the key before the app sees it.

A Mac user with two or more Spaces presses Ctrl+2 as the in-app help ('Ctrl+1 ... Ctrl+9 on macOS') advertises: macOS switches to Desktop 2 instead of tab 2, and the app never receives the keydown. The advertised default shortcut silently does nothing in-app for these users; only rebinding or {count}gt works, and nothing in the UI explains why.

packages/app-core/src/lib/cm-live-preview.ts
```diff
@@ -402,6 +416,19 @@ class LocalImageWidget extends WidgetType {
     image.alt = this.alt
     image.loading = 'lazy'
     image.draggable = false
+    // Obsidian-style size hints (#570). The attribute carries the semantic
+    // size, but the embed class stretches images to width: 100% and
+    // presentational attributes lose to any CSS rule, so the hint also goes
+    // on as inline style. The class's max-width: 100% still caps a hint
+    // wider than the pane.
+    if (this.width) {
```
adibhanna, line 427: **correctness / CONFIRMED**: `LocalImageWidget.toDOM` applies #570 size hints only when truthy onto a cached img element whose cache key (url|mtime) does not include the hint, so a reused element keeps stale width/height attributes and inline styles after the hint is removed or shrunk.

Type `![[pic.png|300x200]]`; the image renders and, once loaded, the element is admitted to the #472 element cache keyed by url|mtime. Put the cursor on the line (widget destroyed, element disconnected but still cached), edit the hint to `|300` or delete it entirely, and move the cursor off: eq() correctly rebuilds the widget, but `takeCachedImage` returns the same element still carrying width=300/height=200 attributes and inline styles, and since this.width/this.height are now undefined nothing clears them, so the image stays at the old size (or renders 300px wide with a stale 200px height, distorting the aspect ratio) until the file's mtime changes or the cache evicts. The same leak fires when one note embeds `![[pic.png|100]]` and another embeds `![[pic.png]]` plain: whichever renders second inherits the 100px style.

packages/app-core/src/lib/cm-vim-pending-input.ts
```diff
@@ -0,0 +1,17 @@
...
+ * True while Vim is mid-sequence and the next keypress belongs to it: after
+ * `f`/`t`/`r` awaiting their target character (`expectLiteralNext`), or with
+ * any partial keys buffered (a count like `2`, a `g` prefix, a `"` register).
+ * App-level single-key shortcuts must yield then, or they steal the operand:
+ * `v f m` opened the context menu instead of jumping to the next `m`, and the
+ * orphaned pending motion then swallowed the first key typed after the menu
+ * closed (#568).
+ */
+export function vimAwaitsNextKey(view: EditorView | null | undefined): boolean {
+  const vim = view ? getCM(view)?.state.vim : null
```
adibhanna: **simplification / CONFIRMED**: The new `vimAwaitsNextKey` helper (plus its test file) duplicates the pre-existing exported `isVimAwaitingArgument` in lib/vim-nav.ts:190 (#147), which implements the identical predicate (`expectLiteralNext || keyBuffer.length > 0`); VimNav.tsx now imports and calls both names for the same vim-pending state in the same handler.

No runtime failure today, but the two copies are one behavioral tweak away from diverging: a future fix to pending-key detection (e.g. also treating inputState.operator as pending) applied to one helper leaves the leader-key guard at VimNav.tsx:1018 and the #568 context-menu guards at VimNav.tsx:1073/1150 plus EditorPane disagreeing about whether vim owns the next key, reintroducing the #147/#568 class of stolen-operand bugs in whichever path kept the stale copy. The #568 change could have imported `isVimAwaitingArgument` (already covered by vim-nav.test.ts) instead of adding a second module, test file, and name for the same logic.

packages/app-core/src/App.tsx
```diff
@@ -696,6 +702,15 @@ function App(): JSX.Element {
         state.setWordWrap(!state.wordWrap)
         return
       }
+      // Alt+1..9 (⌃1..9 on macOS) — jump straight to tab N (#497). Position
```
adibhanna: **style / CONFIRMED**: New comment uses an em dash, violating the house writing rule (never use an em dash in code comments, commits, docs, or UI copy).

The newly added comment `// Alt+1..9 (⌃1..9 on macOS) — jump straight to tab N (#497)` contains an em dash; restructure with a comma, colon, or parentheses.

PR #594, opened 2026-08-14
editor: LaTeX command completion in math regions

flokchvtr:
## What

LaTeX command completion inside math regions. Typing `\su` in `$…$`, `$$…$$`, or a ```math fence pops KaTeX commands with the **rendered symbol** in the completion row's icon slot — same row layout as the slash commands.

- ~250 curated commands: greek, big operators, accents, fonts (`\mathbb`, `\mathcal`, …), relations, arrows, named functions, delimiters, and matrix/cases/aligned environments.
- Argument-taking commands insert as **snippets**: accepting `\frac` lands the cursor in the numerator, Tab moves to the denominator. Big operators scaffold their usual bounds — `\sum` inserts `\sum_{i=1}^{n}` with each bound selected in turn, so the defaults are one Tab away from being kept or one keystroke from being replaced.
- Environments unfold on their own lines (`\pmatrix` → `\begin{pmatrix}…\end{pmatrix}`).

## How

One more completion source in the editor's existing autocompletion stack, plus a `_kind === 'latex'` branch in the shared `renderCompletion` (the pattern callouts and wikilinks already use). No new dependencies — `katex` and `@codemirror/autocomplete` are already in the tree.

Math-region detection counts **unmatched** `$`/`$$` delimiters rather than closed pairs: a formula still being typed (no closing `$` yet) already counts as math, which is exactly when completion matters. Escaped dollars (`\$5`) and code regions are excluded; a fence whose info string is `math` is treated as a math region in its own right, matching how remark-math renders it.

## Tests

`cm-latex-completions.test.ts` covers the context detection (inline open/closed, block open/closed, ```math vs other fences, escaped dollars, inline code) and the token matcher (`\su`, bare `\`, `\\` row-break rejection). `npm run typecheck` and the app-core suite pass.

## Out of scope

Completion only activates in math regions, so `\` stays inert in prose. The command table is a curated constant — happy to trim or extend it to taste.

### DIFF
--- a/packages/app-core/src/components/EditorPane.tsx
+++ b/packages/app-core/src/components/EditorPane.tsx
@@ -120,6 +120,7 @@ import { mathBlockArrowKeymap } from '../lib/cm-math-nav'
 import { slashCommandSource, slashCommandRender } from '../lib/cm-slash-commands'
 import { calloutTypeSource } from '../lib/cm-callouts'
 import { dateShortcutSource } from '../lib/cm-date-shortcuts'
+import { latexCommandSource } from '../lib/cm-latex-completions'
 import { wikilinkSource, wikilinkHeadingSource, atNoteSource } from '../lib/cm-wikilinks'
 import { linkRangeAtCursor, markdownLinkAt } from '../lib/internal-links'
 import { setBlockType, toggleWrap, wrapLink } from '../lib/cm-format'
@@ -1779,6 +1780,7 @@ export function EditorPane({ pane }: { pane: PaneLeaf }): JSX.Element {
               slashCommandSource,
               calloutTypeSource,
               dateShortcutSource,
+              latexCommandSource,
               atNoteSource,
               hashtagSource,
               wikilinkSource,
--- a/packages/app-core/src/lib/cm-latex-completions.test.ts
+++ b/packages/app-core/src/lib/cm-latex-completions.test.ts
@@ -1,0 +1,79 @@
+import { describe, expect, it } from 'vitest'
+import { EditorState } from '@codemirror/state'
+import { markdown } from '@codemirror/lang-markdown'
+import { isInMathContext, latexTokenBefore } from './cm-latex-completions'
+
+function state(doc: string): EditorState {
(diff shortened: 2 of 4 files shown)


adibhanna:
Merged into the v2.28.2 release branch (8cd147d) and shipping in 2.28.2. Thank you, this is a genuinely nice piece of work: the rendered symbol in the icon slot is the part that makes it feel finished, and the unmatched-delimiter reasoning is exactly right, since the closing `$` is precisely what does not exist yet while you are typing.

I drove it in the app before merging: inside `$$…$$`, typing `\su` opens 13 options with `\sum` boosted first, Enter lands `\sum_{i=1}^{n}` with the bounds a Tab apart, and the same keystrokes in prose stay inert.

Three review findings landed in a follow-up commit (5eb6d5c) rather than in yours, which is how this repo keeps contributor work intact:

1. **Typst notes.** ZenNotes has a `math_renderer` setting, and a note set to Typst takes different syntax, so `\frac{}{}` would be wrong there every time. The source now reads the renderer facet the pane is configured with and stays silent unless it is KaTeX.
2. **Dollars inside code counted as delimiters.** A note containing `echo $$` in a shell block (the shell's PID) flipped the parity for everything below it, so `\` popped LaTeX commands in plain prose from that point on; a `$5` in inline code did the same for the rest of its line. Both counts now skip whatever the syntax tree says is code, which is the same question your `codeContext` was already asking at the cursor.
3. **Two cost trims.** The block scan started from offset 0 on every `\` typed, allocating the whole prefix as a string; it now looks back a bounded window. And the KaTeX previews are cached between popups, since a bare `\` opens the whole table and every visible row was being typeset again on each open.

None of that changes what you built or how it feels. Tests came along for each: code-fence and inline-code dollars, and the Typst gate.

If you want to keep going in this area, the obvious next step is the same treatment for Typst syntax behind the same setting.

adibhanna:
This was merged by hand into 2.28.2 as `8cd147d` and shipped; closing the PR to match. Thanks again.

PR #595, opened 2026-08-14
feat(editor): frontmatter tag autocomplete and clickable tags

junereycasuga:
This change makes frontmatter `tags:` values behave more like inline `#tags`:

- Typing inside a frontmatter `tags:` field (inline list, scalar, or block list) now suggests existing vault tags.
- Frontmatter tags are now rendered as clickable chips; clicking one opens the tag view.
- Inline `#tag` autocomplete is suppressed inside frontmatter so the two sources don't clash.
- Leading `#` in frontmatter tag values is stripped for both autocomplete and click targets.

SCR-20260814-sgdz

### DIFF
--- a/packages/app-core/src/components/EditorPane.tsx
+++ b/packages/app-core/src/components/EditorPane.tsx
@@ -74,7 +74,7 @@ import {
 } from '../lib/cm-vim-clipboard'
 import { wireYankHighlight, yankHighlightExtension } from '../lib/cm-yank-highlight'
 import { vimVisualHighlightExtension } from '../lib/cm-vim-visual-highlight'
-import { frontmatterStyle } from '../lib/cm-frontmatter'
+import { frontmatterStyle, frontmatterTagExtension } from '../lib/cm-frontmatter'
 import { codeBlockFontPlugin } from '../lib/cm-code-block-font'
 import {
   orderedListRenumber,
@@ -105,6 +105,7 @@ import { hashtagExtension } from '../lib/cm-hashtags'
 import { taskMetadataExtension } from '../lib/cm-task-metadata'
 import { taskRollupExtension } from '../lib/cm-task-rollup'
 import { hashtagSource } from '../lib/cm-hashtag-complete'
+import { frontmatterTagSource } from '../lib/cm-frontmatter-tag-complete'
 import { applyHighlight, HIGHLIGHT_COLORS, highlightExtension } from '../lib/cm-highlight'
 import { wikilinkRenderExtension } from '../lib/cm-wikilink-render'
 import { mathRenderExtension } from '../lib/cm-math-render'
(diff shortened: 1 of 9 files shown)


adibhanna:
Merged into the v2.28.2 release branch (80a90a9) and shipping in 2.28.2. Thank you, this closes a gap that has bothered people for a while: frontmatter tags were already indexed and searchable, they just did not behave like tags where you actually write them.

Two things I want to call out as good judgment on your part. Extracting `frontmatterRange` and having the existing decoration builder use it, rather than adding a second copy of the frontmatter scan, is exactly right for this codebase. And the click handler matching the inline hashtag one byte for byte means the two behave identically, including the mousedown interception so the caret does not land in the tag first.

I verified it in the app before merging: chips render for inline lists, scalars, and block lists, clicking one opens the tag view, completion fires inside a `tags:` field, and inline `#` completion stays out of the way there.

One finding landed in a follow-up commit (a342da2), kept out of yours as this repo does:

**A capital `Tags:` was not treated as the tags field.** `parseFrontmatterFields` in shared-domain lowercases keys, so `Tags:` and `TAGS:` are the tags field everywhere the vault is indexed: those notes appear under their tags in the Tags view, in search, and in the CLI. The editor's inline branch matched `tags` case-sensitively, so a note written with a capital T had tags the rest of the app knew about while the editor would neither chip nor complete them. Interestingly your own block-list branch already lowercased (`key[1].toLowerCase() === 'tags'`), so the two halves disagreed with each other. Both now go through one `frontmatterTagsValue` helper phrased the way the shared parser phrases it.

I also added a test matrix rather than just a case fix, because this feature now has two parsers deciding what a frontmatter tag is: the shared one that gets indexed, and the editor's scan, which has to exist because the shared one returns tags without positions and decorations need offsets. They cannot be merged, so the test asserts that the chips a note renders are exactly what `frontmatterTags` returns for it, across inline lists, scalars, block lists, quoted values, `#`-prefixed values, capitalized keys, and two fields that only look like tags. If either side drifts, it fails.

Nice work, and the screenshot in the description made the review much easier.

adibhanna:
Following up on my own claim above: when I wrote it, the inline-`#` suppression had only been covered by your unit tests, and my in-app run of the completion had been muddied by the probe editing a seeded note. I went back and exercised all four paths properly in the running app, after the casing fix:

| what | result |
| --- | --- |
| `tags: dra` (lowercase key) | suggests `drafting`, `dragons` |
| `Tags: dra` (capital key) | suggests the same, which is what the follow-up fixed |
| `note: #dra` inside frontmatter | no popup, your suppression works |
| `#dra` in the body | popup, inline completion unaffected |

One reading in between showed "no popup" for the lowercase case and briefly looked like a regression. It was my probe clicking before the editor had focus: repeating the same case at the end of the run completed normally. Flagging it only so the record is accurate, since I had already told you it was verified before it fully was.

Nothing changed in the code as a result. Your work and the follow-up both stand as merged.

junereycasuga:
@adibhanna since this is already added in v2.28.2 and it's already released, shall we close this?

junereycasuga:
closing after the changes merged into the v2.28.2 release branch

PR #602, opened 2026-08-15
fix(editor): focus first row when inserting a table via /table

### LINKED ISSUE #663: /table insert leaves focus inside the first header cell instead of after the table widget
**Description:**
When inserting a table via the /table slash command, the editor focus is moved into the first header cell of the newly inserted table. The expected behavior is for focus to land after the table block widget, so the user can continue typing in the body of the note without first escaping the table.

**Steps to reproduce**
1. Open any note in ZenNotes.
2. Type /table and select the table slash command.
3. Observe where the cursor/focus ends up.

**Expected behavior**
Focus should be placed immediately after the newly inserted table block widget, leaving the table empty but not active.

**Actual behavior**
Focus is placed inside the first header cell of the table, requiring an extra key interaction (e.g. Esc or arrow navigation) to exit the table and continue writing.

**Additional context**
This affects the keyboard-first flow because /table interrupts note writing instead of acting like a passthrough block insert.

junereycasuga:
Fixes https://github.com/ZenNotes/zennotes/issues/663

When inserting a table with `/table`, focus now lands in the first header cell instead of being left after the block widget.

**Changes:**
- Added `focusTableCell(view, tableFrom, row, col)` helper in `cm-table.ts`.
- Updated `/table` slash command in `cm-slash-commands.ts` to focus the first header cell once the widget renders.
- Updated test in `cm-slash-commands.test.ts` to assert the first header cell is focused.

**Tested:**
- `npm run test:run --workspace @zennotes/app-core -- cm-slash-commands.test.ts`
- `npm run typecheck --workspace @zennotes/app-core`

**Before:**

https://github.com/user-attachments/assets/7c8afc7a-9be8-4a42-9d70-8b1a1f42c9f9

**After:**

https://github.com/user-attachments/assets/9c4e6250-0f09-44b7-bfb9-72797f0fd297

### DIFF
--- a/packages/app-core/src/lib/cm-slash-commands.test.ts
+++ b/packages/app-core/src/lib/cm-slash-commands.test.ts
@@ -1,6 +1,7 @@
 // @vitest-environment jsdom
 
 import { CompletionContext } from '@codemirror/autocomplete'
+import { markdown, markdownLanguage } from '@codemirror/lang-markdown'
 import { EditorState } from '@codemirror/state'
 import { EditorView } from '@codemirror/view'
 import { describe, expect, it, vi } from 'vitest'
@@ -14,6 +15,7 @@ import {
   templateSlashCommandSource,
   blockInsertPadding
 } from './cm-slash-commands'
+import { tablePlugin } from './cm-table'
 
 type Source = typeof templateSlashCommandSource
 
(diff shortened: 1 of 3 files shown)


junereycasuga:
Closing after getting fixed on the v2.36.0 release branch in [24448dc](https://github.com/ZenNotes/zennotes/commit/24448dc).

PR #604, opened 2026-08-15
editor: Add Typst command completion in math regions

flokchvtr:
Following up on #594 and the feedback from the 2.28.2 merge, this PR brings the same math auto-completion experience to Typst syntax.

Because Typst writes math differently than LaTeX (parentheses instead of backslashes and braces), I built a dedicated completion source for it. The two sources are mutually exclusive: they check `mathRendererOf`, so you'll only get suggestions relevant to the active math renderer.

Completion fires after two letters to avoid popping up on single-letter math variables (though manual trigger still works for a single letter). For math-region detection, I reused `isInMathContext` this time.

The completion table covers: snippets (frac, sum, integral, lim, mat, cases...), accents, text styles (bold, upright, cal, bb), Greek letters, comparisons, sets, and the usual operators. It also includes arrows with their ASCII shorthands (e.g., `arrow.r` / `->`).
*Note: `plus.circle` and `times.circle` are deprecated, so the table uses the canonical `plus.o` and `times.o` instead.*

Following the LaTeX review, previews simply use exact Unicode characters (like α, ∑, ℝ) instead of a full render whenever possible to keep things lightweight. Snippets with arguments still use the existing shared, cached render queue.

Tests cover:

* The token matcher (handling the two-letter minimum and dotted names like `arrow.l.r.double`).
* The typesetter gate (ensuring LaTeX and Typst don't step on each other's toes).
* Ensuring popups stay out of standard prose and code blocks.
* Verifying that every snippet preview is well-formed and doesn't leak syntax into the compiler.

### DIFF
--- a/packages/app-core/src/components/EditorPane.tsx
+++ b/packages/app-core/src/components/EditorPane.tsx
@@ -123,6 +123,7 @@ import { slashCommandSource, slashCommandRender } from '../lib/cm-slash-commands
 import { calloutTypeSource } from '../lib/cm-callouts'
 import { dateShortcutSource } from '../lib/cm-date-shortcuts'
 import { latexCommandSource } from '../lib/cm-latex-completions'
+import { typstCommandSource } from '../lib/cm-typst-completions'
 import { wikilinkSource, wikilinkHeadingSource, atNoteSource } from '../lib/cm-wikilinks'
 import { linkRangeAtCursor, markdownLinkAt } from '../lib/internal-links'
 import { setBlockType, toggleWrap, wrapLink } from '../lib/cm-format'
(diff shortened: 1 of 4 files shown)


adibhanna (approved):
Reviewed against v2.29.0. The completion source is renderer-gated, math-context-scoped, and covered by focused tests. Verified the integrated tree with app-core typechecking, all 1,648 app-core tests, direct compilation of all 25 snippet previews and 120 completion labels using the bundled Typst compiler, and the fresh four-platform CI matrix. No blocking findings.

adibhanna:
Reviewed against `v2.29.0` and merged in `051be0015d0baae1d2a10c56dbc314146e214f2a`.

Verification completed:
- `@zennotes/app-core` TypeScript check passed
- 1,648 app-core tests passed (1 skipped)
- All 25 snippet previews and 120 completion labels compiled with the bundled Typst compiler
- Fresh CI passed on macOS, Windows, Ubuntu x64, and Ubuntu ARM

No blocking findings.

PR #707, opened 2026-08-30
Fix(editor): arrow keys navigate the @, / and [[ menus again

ArditZubaku:
## What

Arrow keys navigate an open completion menu again — the `@` date/note suggestions, the `/` slash menu, the `[[` reference picker and the callout `[!` type picker. Before this, ↑/↓ moved the caret instead, and the caret move closed the menu, so the only way to pick an item was Ctrl+N/Ctrl+P.

## Why it broke

`autocompletion({ defaultKeymap: false })` skips the stock completion keymap so mac AltGr-style layouts keep their `` Alt-` ``/`Alt-i` characters. The filtered replacement (`completionKeymapForEditor`) was re-added by spreading it into the editor's general `keymap.of([...])`.

Upstream mounts that keymap at `Prec.highest`. The spread put it at default precedence and **last** in the array — and CodeMirror runs same-key bindings in array order (`buildKeymap` pushes onto `binding.run`). So `defaultKeymap`'s `ArrowUp`/`ArrowDown` (`cursorLineUp`/`cursorLineDown`) ran first, returned `true`, and the completion never saw the key.

`EditorPane` and `PinnedReferencePane` spread it last, so they were broken. `QuickCaptureApp` and `TemplateEditorModal` happened to spread it first, so they worked — which is why the behaviour differed between the main editor and the Quick Note window.

## The fix

A new `completionKeymapExtension` mounts the same filtered bindings at `Prec.highest`, matching what `@codemirror/autocomplete` does for its own keymap. All four editors mount it next to `completionNavKeymap` instead of inlining the spread, which removes the ordering trap for future edits.

No behaviour change for the AltGr fix — the two mac chords stay filtered out.

## Testing

- New `cm-completion-nav-arrows.test.ts`: ↓↓↑ moves the highlighted option, the menu stays open and the caret does not move; with no menu open the arrows fall through to the next binding.
- Verified the old wiring fails that test and the new wiring passes.
- `npm run typecheck`, `npm run test:run` (1827 passed) and `npm run build` all pass locally on Node 22.

## Before / After

**Before** — ↓ moves the caret and the `@` menu closes.

https://github.com/user-attachments/assets/199e7e7a-d3f5-4c79-805f-731e6774e78a

**After** — ↓/↑ move the highlighted item, the menu stays open.

https://github.com/user-attachments/assets/d77264d0-4d32-4138-9940-a3a991c86225

### DIFF
--- a/packages/app-core/src/components/EditorPane.tsx
+++ b/packages/app-core/src/components/EditorPane.tsx
@@ -62,7 +62,7 @@ import { forwardOnCheckboxArrow } from '../lib/cm-forward-task'
 import { markerHopCommands } from '../lib/cm-marker-hop'
 import { isInMarkdownCode } from '../lib/cm-auto-pairs'
 import { toggleCheckbox } from '../lib/cm-toggle-checkbox'
-import { completionKeymapForEditor, completionNavKeymap } from '../lib/cm-completion-nav'
+import { completionKeymapExtension, completionNavKeymap } from '../lib/cm-completion-nav'
 import { vimAwareDefaultKeymap, vimAwareMarkdownKeymap, vimAwareSearchKeymap } from '../lib/cm-vim-default-keymap'
 import { isVimAwaitingArgument } from '../lib/vim-nav'
 import { toCodeMirrorKey, vimHalfPageKeymap } from '../lib/vim-half-page-keymap'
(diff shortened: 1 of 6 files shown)


adibhanna:
Merged into the v2.45.0 release branch, thank you. Verified over CDP in the built app with Vim on and off: the arrows move the highlight in the [[ picker, it stays open, Enter inserts, and Esc still closes it without leaving insert mode. This also closes #739, whose reporter hit exactly this.

